PR #1212, opened 2026-03-15
[FIX] Prompt 123: fix migration order bug and ::text casts

ronniepinnell:
Fix two bugs from code review: (1) Remove FK blocks from 20260320000001 that would abort on fresh schema — FKs reference player_uuid columns added by 20260320000002 which runs AFTER, causing abort. (2) Add ::text cast to all 7 UPDATE backfills in 20260320000002 for type safety. DEV unaffected (migrations applied interactively). Fixes fresh deployment path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced the onboarding setup finish action to ensure all progress is properly saved, completion states are accurately tracked, and the workflow reliably transitions to completion.
  * Improved data integrity for player identifier management to prevent synchronization inconsistencies and ensure consistent records across the platform.


ronniepinnell:
Added docstring comment to add_player_uuid_fkeys() clarifying that fact_goals/fact_shots/fact_penalties are intentionally excluded (role-specific player columns, not gaps). Karen validation complete.

Issue #2936, opened 2026-04-12
#2769-T2b Fix test_auth_cache.py supabase import shadowing in CI


ronniepinnell:
## Goal
Fix the 3 failing tests in `tests/unit/test_auth_cache.py` that break CI on develop.

## Root Cause
The local `supabase/` directory (containing migrations) is treated as a Python namespace package in CI, shadowing the `supabase` pip package. When tests do `from supabase import create_client`, Python resolves to the local directory instead of the pip package.

## Fix Options (pick simplest)
1. Add `__init__.py` guard or `conftest.py` with `sys.path` manipulation to ensure pip package is found
2. Change the mock targets in `test_auth_cache.py` to patch the actual import location (`utils.supabase_client.create_client`)  
3. Add a `py.ini` or `pyproject.toml` config to exclude `supabase/` from Python path resolution

## Steps
1. Reproduce the failure in a local environment
2. Identify root cause
3. Apply the fix
4. Verify fix passes all relevant tests

## Acceptance Criteria
- [ ] `pytest tests/unit/test_auth_cache.py` passes locally
- [ ] CI green on develop after merge

## Context
- Epic: #2769 (E0051)
- Blocking task: #2786
- CI logs: 3 failed, 518 passed, 2 skipped in API Tests job
- Error: `AttributeError: <module 'supabase' (namespace)> does not have the attribute 'create_client'`

**Labels:** priority:p0, type:bug, area:factory, factory:task, lane:product, autonomy:green, model:sonnet, machine:any, milestone:m0002

ronniepinnell:
## Gatekeeper Report

| ✅ | **triage** | approve | Warning: Missing `size:*` label. |
| ✅ | **plan** | approve | No spec-ref found (§0.D12 DoR requires one).; No `## Outcome` section — how will we know this shipped successfully?; No `## Required Reading` section — planner should add one.; No guardrails section — rules/ injection will cover base rules but issue-specific guardrails are missing. |
| ✅ | **plan-review** | approve | Review passed — ready for factory. Notes: §17: issue touches schema — ensure a migration file exists before code. |

**Complexity:** M

ronniepinnell:
Claimed by `daemon-v15-mothership` on `mothership`.

ronniepinnell:
**Node `mothership` completed task.**

| Metric | Value |
|:---|:---|
| Status | COMPLETED |
| Duration | 215s |
| Model | `sonnet` |
[...]

PR #2947, opened 2026-04-12
[FIX] #2936 — prevent supabase/ dir from shadowing pip package in CI

ronniepinnell:
## Summary

Fixes the 3 failing tests in `api/tests/unit/test_auth_cache.py` that were breaking CI on develop. Root cause: `api/__init__.py` causes pytest to add the repo root to `sys.path` (so `api.*` can be imported as a package). The repo root contains a `supabase/` migrations directory with no `__init__.py`, which Python silently treats as a namespace package — shadowing the real `supabase` pip install. When tests call `patch("supabase.create_client", ...)`, the patch mechanism calls `getattr(namespace_package, 'create_client')`, which raises `AttributeError` before the test even runs.

Fix: in `api/tests/conftest.py`, push the repo root to the END of `sys.path` (so site-packages wins the `supabase` lookup) and evict any stale namespace cache entry before tests run.

## What's in the diff

- `api/tests/conftest.py` — 17-line addition at the top: `sys.path` reordering + `sys.modules.pop("supabase", None)` to clear any pre-cached namespace entry.

## Verification

- [x] `python -m pytest tests/unit/test_auth_cache.py -v` — 18/18 passed (was 3 failing, 15 passing)
- [x] `python -m pytest tests/unit/ -v` — 521 passed, 2 skipped, 0 failures (no regressions)
- [x] `python -m py_compile api/tests/conftest.py` — no syntax errors

## Out of scope

- Duplicate migration timestamps (pre-existing, unrelated to this fix)
- Future-dated migrations (pre-existing, unrelated)

## Test plan

- [ ] Merge to `develop` and confirm `API Tests` CI job goes green
- [ ] Verify no other unit test files regress

## Summary by CodeRabbit

* **Chores**
  * Improved test setup to ensure proper package resolution order and dependency loading during test execution.


ronniepinnell:
Superseded by #3129 (merged) — bundled the supabase singleton test split with Ruff grep + zombie workflow fixes.

PR #9330, opened 2026-04-29
[FCT003E] E06: JSONL Session Sync

ronniepinnell:
## Summary

Parse Claude Code JSONL session files and sync turn-level data to qa.conversation_turns. Post-session hook (primary) + companion reconciliation (secondary).

## Epic
BEN-1714: [FCT003E] E06: JSONL Session Sync

## Tasks Included
- BEN-1748: Build Claude Code JSONL parser

## Test Results
```
tests/milestones/fct003e/test_fct003e_acceptance.py::TestJSONLSync
  PASSED  test_post_session_hook_parses_jsonl
  SKIPPED test_companion_reconciliation_catches_missed (integration, deferred)
  PASSED  test_r2_offload_enforced
  SKIPPED test_interactive_session_has_turns (live-DB, deferred)

tests/infra/test_session_parser.py
  38 passed in 0.98s
```

## Acceptance Criteria
- Both interactive and factory sessions have turn-level data. R2 offload works. Companion catches gaps.

## Changes
- `scripts/infra/session_parser.py` — new module: discover_sessions() + parse_session()
  - Parses Claude Code JSONL session files (~/.claude/projects/*/*.jsonl)
  - Extracts user/assistant/tool_use/tool_result turns
  - Computes per-turn cost from token_parser._PRICING
  - Flags content >4KB for R2 offload (r2:// placeholder sentinel)
  - Returns list[dict] matching qa.conversation_turns schema
- `tests/infra/test_session_parser.py` — 38 unit tests (TDD)
- `tests/milestones/fct003e/test_fct003e_acceptance.py` — TestJSONLSync implemented (2 live, 2 deferred)
- `docs/MASTER_INDEX.md` — updated to reference session_parser.py
- `.claude/prompts/4632_build_claude_code_jsonl_parser.md` — factory prompt for BEN-1748

## Known Limitations
- Cache token costs (cache_read/creation_input_tokens) not included — _PRICING has no cache rates. Documented in _compute_cost docstring.
- Companion reconciliation and live-DB tests deferred to E06 integration phase.

## Summary by CodeRabbit

* **New Features**
  * Added JSONL session parsing capability to transform Claude Code session data into structured conversation turns.
  * Implemented automatic content offloading for large messages to external storage.
  * Added cost computation for parsed sessions based on token usage.
  * New session discovery and parsing functionality with schema validation.


ronniepinnell:
## Code Review (claude-sonnet-4-6)

**Verdict:** APPROVED

**Blockers:** 0
**Warnings:** 3
**Suggestions:** 4

---

### Summary

`session_parser.py` is well-structured and correctly implements JSONL parsing, R2 offload, and cost routing. The test suite has strong coverage of the primary happy paths. Three warnings require attention before this becomes load-bearing in production.

---

### Warnings

**W1 — Cache token cost omission is undocumented at the call site (only in the docstring)**

`_compute_cost()` silently ignores `cache_read_input_tokens` and `cache_creation_input_tokens`. The docstring documents this, but the cost field returned is indistinguishable from a full-cost figure. Downstream consumers reading `cost_dollars` from `qa.conversation_turns` have no signal that the value is understated. Add a `cost_is_partial` boolean column or a `cost_notes` text column to the row dict (set to `"cache_excluded"` when either cache field is nonzero) so dashboards can flag partial values.

**W2 — `_PRICING` is imported from a private name (fragile coupling)**

`from scripts.infra.token_parser import _PRICING` imports a module-private symbol. If `token_parser` is ever refactored the import silently falls back to `None` cost for every row — the `ImportError` guard catches only import failures, not attribute removal. Promote `_PRICING` to a public `PRICING` dict in `token_parser.py`, or expose a thin `cost_for_model(model, input_tokens, output_tokens)` function so `session_parser` has a stable contract.

**W3 — No path traversal guard on `session_id`**

`session_id = jsonl_path.stem` is used unvalidated as part of the `r2://` sentinel stored in the NOT NULL `content` column and in `content_r2_path`. A crafted filename such as `../../etc/passwd.jsonl` will produce a sentinel `r2://../../etc/passwd/0` which, when the caller uses it to build an actual R2 key, becomes a path traversal. Validate with `re.fullmatch(r'[\w\-]+', session_id)` or `uuid.UUID(session_id)` and reject/sanitize before use.

---

### Suggestions

**S1 — Empty assistant text row emitted unconditionally**

`_make_assistant_rows()` always emits the assistant text row even when `combined_text == ""` (tool-only assistant turns). The comment says "the turn happened" but an empty-content row adds noise and wastes the NOT NULL column sentinel unnecessarily. Consider skipping the text row when `text_parts` is empty and there are tool_use blocks, or at minimum set `content = "(no text)"` so consumers can filter easily.

**S2 — `_maybe_offload` uses byte length of UTF-8 but threshold name implies chars**

The constant is named `_R2_THRESHOLD_BYTES` and the implementation correctly uses `len(content.encode("utf-8"))` — this is correct. However the acceptance test uses `"y" * 5000` (5000 ASCII chars = 5000 bytes) and the unit test uses `"x" * 5000` — both safely above 4096. Add one test at exactly 4096 bytes and one at 4097 bytes (boundary value) to pin the strict `>` vs `>=` behavior.

**S3 — No test for `discover_sessions` sort order**

`discover_sessions` sorts by `st_mtime` (oldest first) but no test asserts this ordering. Callers relying on chronological sync order will silently get wrong order if the sort is ever changed. Add a test that writes two files with different mtimes and asserts the returned list is oldest-first.

**S4 — `tool_input` stored as raw Python object, not serialized JSON**

`_base_row` stores `tool_input=block.get("input")` as a raw dict. If the caller inserts this directly into Supabase via the Python client it will serialize correctly, but if any code path stringifies the row (e.g. `json.dumps(row)` for logging or queuing) the dict survives fine. The concern is the sibling field `content` for the same turn is `json.dumps(block.get("input", {}))` — the same data exists in two fields with different types (str vs dict). Document this dual representation explicitly or unify: store `tool_input` as a dict everywhere and make `content` the canonical serialized form.

---

### What is correct

- JSONL type dispatch (user / assistant / tool_use / tool_result / skip-all-others) is correct and defensive.
- R2 threshold logic is correct: `>` (strictly greater than) 4096, sentinel stored in content, `content_r2_path` set to identical value.
- Cost formula `(input / 1M) * price_in + (output / 1M) * price_out` matches `token_parser._calculate_cost` exactly.
- `thread_uuid` is fresh per file, `session_id` is the filename stem — both correct per the spec.
- `turn_index` is global across all row types in the file (not reset per message type) — correct for the schema.
- Malformed JSON and non-dict JSON are gracefully skipped with debug logging.
- OSError on file read returns `[]` without crashing.
- All 13 required `qa.conversation_turns` keys present on every row.
- `cites` is always `[]`, `machine` and `issue_number` always `None` — correct per spec (caller fills these).
- Test class `TestJSONLSync` in the acceptance file correctly covers `parse_session` happy path and R2 threshold. Deferred integration tests are appropriately marked `pytest.skip`.

ronniepinnell:
All warnings and suggestions addressed in commit 1be8051c:

**W1 (cache tokens ignored):** `compute_cost()` now receives `cache_read_tokens` and `cache_creation_tokens` from `usage` dict. Cache read billed at 10% of input rate, cache creation at 125%.

**W2 (private `_PRICING` import):** Extracted `compute_cost()` as a public function in `token_parser.py`. `session_parser` now imports and calls it instead of accessing `_PRICING` directly. `_calculate_cost()` also delegates to `compute_cost()` to eliminate the duplication.

**W3 (session_id path traversal):** Added `_sanitize_session_id()` — strips any character that is not a word char or hyphen using `re.sub(r"[^\w\-]", "_", raw)`, with a warning log on sanitization.

**S1 (empty tool-only turns):** Assistant turns with only `tool_use` blocks now set `content = "(no text)"` so consumers can distinguish a tool-only turn from an empty response.

**S2 (boundary value tests):** Added `TestR2BoundaryValues` — one test at exactly 4096 bytes (no offload) and one at 4097 bytes (offloaded).

**S3 (sort order test):** Added `TestDiscoverSessionsSortOrder::test_sorted_oldest_first` — writes two files with a 50ms mtime gap and asserts oldest-first ordering.

**S4 (dual representation comment):** Added inline comment explaining that `content` holds JSON-serialised string (subject to R2 offload) and `tool_input` holds raw dict for JSONB insertion.

All 41 tests passing.

ronniepinnell:
Addressed both CodeRabbit findings in commit 2f048f37:

**Content normalization in `_make_assistant_rows`:** Changed `message.get("content", [])` to:
```python
raw_content = message.get("content")
content_blocks = raw_content if isinstance(raw_content, list) else []
```
This guards against `None` (key exists with null value) and non-list types, both of which would raise `AttributeError` in the for loop.

**Deterministic `thread_uuid`:** Replaced `uuid.uuid4()` with `uuid.uuid5(uuid.NAMESPACE_URL, session_id)`. Re-parsing the same session file now produces the same `thread_uuid`, enabling idempotent re-syncs. Different session files still produce different UUIDs (different `session_id` → different uuid5).

All 41 tests still passing.

PR #9363, opened 2026-04-29
[APP001C] E02: RBAC Gate Unification

ronniepinnell:
## Summary
Remove the legacy `dim_user_roles` RBAC gate from admin layout and unify on `fact_user_role` + `dim_platform_roles`. Users provisioned only in the new RBAC system are no longer silently blocked from admin access.

## Epic
BEN-1937: [APP001C] E02: RBAC Gate Unification

## Tasks Included
- BEN-1952: Replace dim_user_roles gate with fact_user_role in admin layout
- BEN-1955: Migration to backfill legacy RBAC users
- BEN-1953: BEN-1937-REVIEW Exit gate verification

## Test Results
9/9 acceptance tests passing:
- Admin layout does not query dim_user_roles ✓
- Admin layout uses getAdminRole() ✓
- getAdminRole() queries fact_user_role + dim_platform_roles ✓
- Layout denies viewer role ✓
- Backfill migration exists ✓
- Migration inserts into fact_user_role ✓
- Migration maps legacy roles correctly ✓
- Migration skips existing users ✓
- No dim_user_roles gate references in admin layout ✓

## Acceptance Criteria
- [x] Admin layout gates on fact_user_role, not dim_user_roles
- [x] User with only fact_user_role entry can access admin
- [x] Migration backfills legacy users
- [x] Single RBAC gate — no silent access denials

---

## Summary by CodeRabbit

* **New Features**
  * Added an automatic backfill to migrate legacy role assignments into the unified RBAC store.

* **Tests**
  * Added acceptance tests validating admin gate behavior, backfill migration presence, legacy-role mappings, and absence of legacy role queries.

* **Refactor**
  * Replaced legacy authorization checks in admin areas and delegated access decisions to the new RBAC service.


ronniepinnell:
## Code Review: RBAC Gate Unification (BEN-1937)

Reviewed 3 changed files + the upstream `get-admin-role.ts` and schema migrations for `fact_user_role` / `dim_platform_roles`.

---

### Security Assessment: PASS (fail-closed)

The admin gate is fail-closed on every path. `getAdminRole()` returns `'viewer'` when:
- No `orgId` provided
- No authenticated user
- No `fact_user_role` rows found
- No `user_org_membership` fallback found
- Unknown `display_id` not in `DISPLAY_ID_MAP`

The layout redirects on `'viewer'`. No fail-open risk identified.

---

### Issues Found

**[MEDIUM] Silent row drop on unmapped legacy roles — backfill migration line 18-22**

The `CASE` statement has no `ELSE` clause:
```sql
CASE dur.role
  WHEN 'Admin' THEN 'org_admin'
  WHEN 'Coach' THEN 'coach'
  WHEN 'Viewer' THEN 'spectator'
END
```

If any legacy user has a role value outside these three (e.g. `'Owner'`, `'Manager'`, a mixed-case variant like `'admin'`, or any typo), the `CASE` returns NULL, the JOIN to `dim_platform_roles` produces no match, and that user is silently skipped. This is not a security issue (fail-closed), but it is a **data integrity** concern -- you could lose legitimate users during backfill with no warning.

**Recommended fix:** Either add `ELSE 'spectator'` as a safe default, or add a preceding diagnostic query:
```sql
-- Before INSERT: flag unmapped roles
DO $$
BEGIN
  IF EXISTS (
    SELECT 1 FROM public.dim_user_roles
    WHERE role NOT IN ('Admin', 'Coach', 'Viewer')
      AND user_id IS NOT NULL AND org_id IS NOT NULL
  ) THEN
    RAISE WARNING 'dim_user_roles contains unmapped role values — check before proceeding';
  END IF;
END $$;
```

**[LOW] `as any` casts in get-admin-role.ts lines 44-48**

The Supabase joined relation `dim_platform_roles` is cast to `any` twice. This disables type checking on `hierarchy_level`, `is_employee`, and `display_id`. Not a blocker, but a typed interface would catch schema drift at compile time.

---

### Test Coverage Assessment

9 tests across 3 classes. Coverage is structural (file-content assertions) rather than behavioral, which is acceptable for a migration + layout refactor where the real RBAC logic requires a running Supabase instance. The tests verify:

- Layout no longer references `dim_user_roles` (2 tests, one via content check, one via grep)
- Layout calls `getAdminRole()` and checks for `'viewer'`
- `getAdminRole()` queries `fact_user_role` + `dim_platform_roles`
- Backfill migration exists, targets correct tables, maps all 3 roles, uses `NOT EXISTS`

**Missing test:** No test verifies that `getAdminRole()` returns `'viewer'` as its default (the fail-closed contract). Consider adding a test asserting the function signature returns `Promise<AdminRole>` and that `'viewer'` appears as the default return.

---

### Migration Safety: PASS

- `INSERT ... SELECT` is idempotent via `NOT EXISTS` -- safe to re-run
- `user_role_key` auto-generated by existing trigger -- no NOT NULL violation
- `display_id` is nullable on `fact_user_role` -- no issue with omission
- No destructive DDL (no DROP, no ALTER)
- Migration timestamp (`20261102000001`) is sequentially correct after existing migrations

---

### Summary

| Category | Verdict |
|----------|---------|
| Security (fail-open risk) | PASS |
| Migration safety | PASS |
| Data integrity | MEDIUM (unmapped roles silently dropped) |
| Type safety | LOW (as any casts) |
| Test coverage | Adequate for scope |
| Code quality | Good -- clean, well-commented |

**Recommendation:** Address the unmapped-role silent drop (MEDIUM), the rest is good to merge.

ronniepinnell:
Fixed in 3a5bae2a: Added diagnostic DO block that warns if unmapped role values exist in dim_user_roles before the backfill INSERT runs. Addresses the MEDIUM finding about silent row drops.

ronniepinnell:
Fixed in 2b9a2f5a: dim_user_roles.user_id is TEXT (e.g. PL0000000001), not a UUID. Now joins through dim_user to resolve auth_id (UUID) for fact_user_role.user_uuid. Addresses Codex P1 finding.

ronniepinnell:
Addressed both CodeRabbit findings in 1cb0fc15:

1. **Migration team scope**: Added `AND fur.team_uuid IS NULL` to NOT EXISTS check — org-wide backfill rows won't be skipped when a team-scoped row exists for the same user/role/org.

2. **Test precision**: Replaced subprocess grep with in-process checks. Assertions now match executable patterns (`getAdminRole(`, `=== 'viewer'`, `.from('fact_user_role')`) instead of generic substrings.

9/9 tests passing.

PR #9382, opened 2026-04-30
[APP001D] E02: API & Test Fixes

ronniepinnell:
## Summary
Fix API behavioral issues and test infrastructure bugs discovered during E2E walkthroughs.

## Epic
BEN-2119: [APP001D] E02: API & Test Fixes

## Tasks Included
- BEN-2039: Season delete → soft delete (is_active=false, deactivated_at)
- BEN-2099: Player API respects org field config instead of hardcoding medical_notes/jersey_size
- BEN-2064: Smoke test _api_available() → session-scoped fixture that fails loudly

## Test Results
8/8 acceptance tests passing (tests/milestones/app001d/test_app001d_acceptance.py)

## Acceptance Criteria
- Season delete is soft (is_active=false) ✓
- Player API respects org config ✓
- Smoke tests don't skip silently ✓

---

## Summary by CodeRabbit

* **New Features**
  * Seasons can be deactivated (soft-deleted) and are reactivated when updated.

* **Improvements**
  * Player field visibility (e.g., medical notes, jersey size) now respects org-level configuration.
  * Admin dashboard shows only active seasons by default.

* **Bug Fixes**
  * Missing player requests now return 404 instead of partial/mock data.

* **Tests**
  * End-to-end test suite now verifies API health at startup and fails fast if the server is unreachable.


ronniepinnell:
## Code Review — PR #9382

Reviewed all 4 changed files. Three issues found, one blocking.

---

### BLOCKING: Soft-delete breaks test re-run idempotency

**File:** `api/routes/league_ops.py` (line 186) + `tests/e2e/app002a/test_api_smoke.py` (line 71)

The `season` fixture cleanup calls `DELETE` (now a soft-delete), setting `is_active=False`. On the next test run, `upsert` finds the existing row (the `existing` check on line 171 does not filter by `is_active`) and updates it — but never resets `is_active=True` or clears `deactivated_at` unless the caller explicitly passes `is_active: true`. The test fixture does not pass `is_active`.

Result: second test run operates on a deactivated season. The `fetchAdminSeasons` UI query (which now filters `.eq('is_active', true)`) will also hide it.

**Fix options (pick one):**
1. Have the `upsert` endpoint always set `is_active=True` and `deactivated_at=None` when updating an existing row (treat upsert as implicit reactivation).
2. Have the test fixture pass `"is_active": true` in the upsert payload.
3. Change the `existing` check to filter `is_active=true` so soft-deleted rows are treated as new inserts.

---

### NON-BLOCKING: Inline import

**File:** `api/routes/league_ops.py` (line 313)

`from datetime import datetime, timezone` is imported inside the function body, but `from datetime import date` is already at module top (line 5). Move `datetime, timezone` to the top-level import to keep imports consistent and avoid per-call import overhead.

---

### NON-BLOCKING: Bare `except Exception: pass` swallows errors silently

**File:** `api/routes/players.py` (lines 499-500)

`_get_player_field_config` catches all exceptions and returns `{}`, which means a misconfigured `config_team` table, a network blip, or a malformed `config_value` JSON will silently show all fields (including sensitive ones like `medical_notes`). At minimum, log the exception so misconfigurations are observable:

```python
except Exception:
    logger.warning("Failed to load player field config for org %s", org_id, exc_info=True)
```

---

### Summary

| Category | Status |
|----------|--------|
| Security | Acceptable (field filtering defaults to showing, not hiding — minor concern noted above) |
| CLAUDE.md compliance | No violations (no client-side aggregation, no iterrows, no god objects) |
| Code quality | Good — mock removal in players.py is a clear improvement |
| Test change | Good design (fail-loudly > silent-skip), but idempotency bug above needs fixing |

ronniepinnell:
Fixed in ebd1b742: (1) Season upsert re-activates soft-deleted rows, (2) datetime import consolidated to module top, (3) bare except replaced with logger.warning in _get_player_field_config.

PR #9389, opened 2026-04-30
[DAT004B] E02: Security Hardening

ronniepinnell:
## Summary
Close all security gaps identified by the security engineer review: constrain global entity INSERT, consolidate user_org_ids(), fix FK conflicts, add RLS to Category E tables.

## Epic
BEN-2146: [DAT004B] E02: Security Hardening

## Tasks Included
- BEN-2159: Constrain INSERT on global entities (dim_player, dim_team, dim_club)
- BEN-2160: Consolidate user_org_ids() — eliminate public vs auth split
- BEN-2161: Fix team_registration FK constraints (NOT NULL vs SET NULL conflict)
- BEN-2162: Add RLS policies for 10 Category E tables

## Migrations
- `20261108000001_constrain_global_entity_insert.sql` — service_role-only INSERT on global entities
- `20261108000002_consolidate_user_org_ids.sql` — drop auth.user_org_ids(), migrate 11 policies to public
- `20261108000003_fix_team_registration_fk_constraints.sql` — ON DELETE RESTRICT for all FKs
- `20261108000004_category_e_rls_policies.sql` — policies for 10 tables with RLS enabled but 0 policies

## Test Results
5/5 SecurityHardening tests SKIPPED (baseline — verified at HARDEN epic with live DB)

## Acceptance Criteria
- Global entity INSERT gated by service_role (not open WITH CHECK true)
- Single user_org_ids() function (public only, auth dropped)
- team_registration FK constraints consistent (ON DELETE RESTRICT)
- All 10 Category E tables have appropriate RLS policies

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened access controls by adding and tightening row-level security policies across multiple tables.
  * Fixed foreign key behavior to prevent accidental deletion of referenced records.

* **Refactor**
  * Consolidated organization-isolation logic for more consistent org-scoped access.
  * Standardized permission patterns (service-role for writes, authenticated org-scoped reads) across datasets.

* **Chores**
  * Added per-user access controls for push subscription records.


ronniepinnell:
## Code Review -- PR #9389: Security Hardening Migrations

Reviewed all 4 migration files. One blocking issue found.

---

### BLOCKING: Migration 3 -- Wrong constraint names will cause failure

**File:** `20261108000003_fix_team_registration_fk_constraints.sql`

The migration drops constraints named `fk_teamreg_division_soft`, `fk_teamreg_season_soft`, and `fk_teamreg_league_soft`. **These names do not exist.** The actual constraint names in the database (set by `20261104000002` and re-applied by `20261106000001_e06_review_fixes.sql`) are:

- `fk_teamreg_division`
- `fk_teamreg_season`
- `fk_teamreg_league`

What will happen at runtime:
1. `DROP CONSTRAINT IF EXISTS fk_teamreg_division_soft` -- silently no-ops (name doesn't exist)
2. `ADD CONSTRAINT fk_teamreg_division` -- **fails** because `fk_teamreg_division` already exists

**Fix:** Change the DROP statements to use the correct names:

```sql
DROP CONSTRAINT IF EXISTS fk_teamreg_division;
DROP CONSTRAINT IF EXISTS fk_teamreg_season;
DROP CONSTRAINT IF EXISTS fk_teamreg_league;
```

---

### Migration 1 -- Looks good

`20261108000001_constrain_global_entity_insert.sql` -- Clean and correct. Properly wrapped in BEGIN/COMMIT. The `DROP POLICY IF EXISTS` + `CREATE POLICY` pattern is safe. Correctly targets `service_role` only for INSERT on global entities, which aligns with the project rule that dim_player/dim_team/dim_club are global (no org_id).

---

### Migration 2 -- Looks good, one note

`20261108000002_consolidate_user_org_ids.sql` -- All 11 policies from the original `20260603000001_rls_policies.sql` that used `auth.user_org_ids()` are accounted for. The two that were on dim_player and dim_team were already dropped by `20261104000005`, leaving exactly 11 to migrate here. The `DROP FUNCTION IF EXISTS auth.user_org_ids()` at the end is safe since all consumers are migrated first within the same transaction.

**SUGGESTION:** The original policies used `= ANY(auth.user_org_ids())` (array operator) while the replacements use `IN (SELECT public.user_org_ids())` (subquery). Both are functionally correct since `public.user_org_ids()` returns `SETOF text`. The `IN (SELECT ...)` form is arguably clearer. No action needed, just noting the pattern change.

---

### Migration 4 -- Mostly good, minor observations

`20261108000004_category_e_rls_policies.sql`

**The DO $$ block for fact_team_venue (lines 71-77):** This will work correctly. The `IF EXISTS` check against `pg_class` with `relnamespace = 'public'::regnamespace` is the right pattern. The `EXECUTE` with string literals is necessary inside a DO block since DDL policy statements can't be conditional otherwise. Correct approach.

**SUGGESTION:** The DO block runs inside the outer BEGIN/COMMIT transaction, but policies created via EXECUTE inside a DO block are still transactional in Postgres, so this is safe. However, if `fact_team_venue` does NOT exist and the block is skipped, you'll have no record that it was skipped. Consider adding a `RAISE NOTICE` in the ELSE branch for operational visibility:

```sql
ELSE
  RAISE NOTICE 'fact_team_venue does not exist, skipping RLS policies';
```

**cv_correction_log (line 37):** The `org_id::text` cast is noted in the comment as needed because org_id is UUID type while `user_org_ids()` returns text. This is correct but worth verifying that the cast doesn't prevent index usage on `org_id`. If this table grows large, consider a functional index on `org_id::text` or changing `user_org_ids()` to return UUID.

---

### Summary

| Migration | Verdict |
|---|---|
| 1 - Constrain global entity INSERT | PASS |
| 2 - Consolidate user_org_ids | PASS |
| 3 - Fix team_registration FKs | **BLOCKING** -- wrong constraint names |
| 4 - Category E RLS policies | PASS (minor suggestions) |

Fix migration 3's constraint names and this is ready to merge.

ronniepinnell:
Fixed in de6c40672: Corrected constraint names from `fk_teamreg_*_soft` to `fk_teamreg_*` — the actual names used in the migration history.

PR #9391, opened 2026-04-30
[DAT004B] E03: Spec/Implementation Alignment (#2147)

ronniepinnell:
## Summary
Aligns actual schema implementation with the spec in `rules/areas/schema.md`. Four changes:
- Renamed `system_key` columns to `{entity}_key` across 112 tables (migration + Python + TypeScript)
- Aligned key prefixes (DP→PL, DG→GM, DN→DV, DS→SC) and digit widths (PL{6D}, TM{5D}, GM{5D}) with spec
- Added UUIDv5 GENERATED ALWAYS to 34 tables with true natural keys; dropped gen_random_uuid() defaults from ~112 ETL-managed tables
- Fixed dim_club: UNIQUE on club_name, switched to generate_club_key(), dropped orphaned sequence, UUIDv5

## Epic
BEN-2147: [DAT004B] E03: Spec/Implementation Alignment

## Tasks Included
- BEN-2163: Rename system_key to {entity}_key across all tables
- BEN-2164: Align key prefixes and digit widths with spec
- BEN-2165: Add UUIDv5 GENERATED ALWAYS to fact/dim tables
- BEN-2166: Fix dim_club: UNIQUE constraint, sequence, UUIDv5

## Test Results
- 66/68 passing (2 pre-existing failures unrelated to this epic)
- 24 acceptance tests skipped (require live DB)

## Acceptance Criteria
- [x] All `system_key` columns renamed to `{entity}_key`
- [x] Key prefixes match spec (PL, TM, GM, SC, etc.)
- [x] Variable digit widths (PL{6D}, TM{5D}, etc.)
- [x] UUIDv5 GENERATED ALWAYS on tables with natural keys
- [x] dim_club: UNIQUE on club_name, proper sequence, UUIDv5
- [x] Python namespace UUID unified with DB

---

## Release Notes

* **Database Schema Updates**
  * Renamed generic key columns to entity-specific keys (e.g., `game_key`, `player_key`, `official_key`) across tables for improved clarity and data integrity.

* **Data Generation**
  * Enhanced key generation with variable precision and improved deduplication logic.

* **UUID Generation**
  * Implemented deterministic UUIDs for improved consistency and data reproducibility.

* **API & UI**
  * Updated endpoints and dashboard components to use entity-specific key fields for highlights and officials management.


ronniepinnell:
## Code Review: [DAT004B] E03 Spec/Implementation Alignment

**Reviewer:** code-reviewer agent
**Scope:** 4 migration files, `key_utils.py`, `key_generator.py`, `supabase_manager.py`, plus spot-checks on remaining files.

---

### CRITICAL: Migration 4 not idempotent — will fail on re-run

`20261109000004_fix_dim_club.sql` line 11:
```sql
ALTER TABLE public.dim_club
  ADD CONSTRAINT dim_club_club_name_key UNIQUE (club_name);
```
This is **not wrapped in an existence check**. If the constraint already exists (e.g., migration re-run after partial failure), it will throw `constraint already exists` and abort the transaction, blocking the TRUNCATE + UUIDv5 conversion below it. Every other migration in this PR uses `DO $$ BEGIN ... EXCEPTION ... END $$` guards — this one should too. Wrap it:
```sql
DO $$ BEGIN
  IF NOT EXISTS (SELECT 1 FROM pg_constraint WHERE conname = 'dim_club_club_name_key') THEN
    ALTER TABLE public.dim_club ADD CONSTRAINT dim_club_club_name_key UNIQUE (club_name);
  END IF;
END $$;
```

Same issue with `DROP CONSTRAINT IF EXISTS dim_club_pkey` followed by `ADD CONSTRAINT dim_club_pkey PRIMARY KEY (club_uuid)` — the ADD is not guarded.

**Severity:** High. Production deploy could fail on retry.

---

### BUG: Stale docstring examples in key_generator.py

`generate_entity_key()` docstring claims:
```python
>>> generate_entity_key('dim_player', 1)
'DP0000000001'
```
But `dim_player` maps to prefix `'PL'` with width 6, so the actual output is `'PL000001'`. The `'DP'` prefix is the OLD prefix that this PR replaces. Same for `fact_events` example — says `'FE0000000042'` but `FE` has default width 10, which happens to be correct, though worth verifying the prefix hasn't changed.

Fix both examples to reflect the new prefixes and widths.

**Severity:** Medium. Misleading docstrings will confuse future contributors.

---

### DESIGN: Python vs SQL UUID generation will produce different values

Python `inject_three_tier_keys()` generates UUIDv5 from **all non-excluded columns** concatenated (line ~1528 of key_utils.py):
```python
key_cols = sorted(c for c in target_df.columns if c not in _EXCLUDE_COLS ...)
natural_keys = table_name + '|' + ... '|'.join(f"{col}={val}" ...)
```

SQL `GENERATED ALWAYS AS STORED` uses a **single natural key**:
```sql
uuidv5_generate(benchsight_namespace(), 'player:' || COALESCE(player_id::text, ''))
```

For Tier 1 tables (dim_player, dim_game, etc.), the Python-generated UUID and the Postgres-generated UUID will **never match**. This is acceptable ONLY because `supabase_manager.py` strips the UUID column before upload (the `GENERATED_ALWAYS_UUID_TABLES` stripping logic). But this creates a subtle correctness risk:

- Any code path that uploads to a Tier 1 table WITHOUT going through `supabase_manager.upload_table()` will get a Postgres error ("cannot insert a non-DEFAULT value into column ... defined as GENERATED ALWAYS").
- The ETL CSV output files will contain the Python-computed UUIDs, which differ from what Supabase stores. Any debugging that compares CSV UUIDs to DB UUIDs will be confused.

**Recommendation:** Document this divergence explicitly in the BENCHSIGHT_NS comment block in `key_utils.py`. Consider aligning the Python generation to use the same single-natural-key formula for Tier 1 tables, so CSV and DB UUIDs match.

**Severity:** Medium (design debt, not a runtime bug given the stripping logic).

---

### MINOR: UUID stripping logic assumes single UUID column per table

In `supabase_manager.py`:
```python
uuid_col = next((k for k in (records[0] if records else {}) if k.endswith('_uuid')), None)
```
This picks the **first** `_uuid` column it finds. If a table has FK UUID columns (e.g., `game_uuid` on `fact_events`), it might strip the wrong one. The logic should match the entity-specific UUID column name, not grab the first `_uuid` it sees.

**Recommendation:** Use `_entity_uuid_column(table_name)` pattern or hardcode `{entity}_uuid` derivation to match the correct column.

**Severity:** Medium-High. Could silently strip FK UUID columns instead of the PK UUID on tables with multiple `_uuid` columns.

---

### MINOR: Migration 2 creates sequences referenced in next_entity_key() but doesn't CREATE them

`next_entity_key()` references sequences like `public.player_key_seq`, `public.team_key_seq`, `public.club_key_seq`, etc. Migration 2 only RENAMES `seq_system_key_*` to `legacy_*_key_seq` — it does not create the new `player_key_seq`, `team_key_seq` etc. If `_seed_key_seq()` from migration 20261104000008 already created them, that's fine, but there's no explicit dependency comment. If those sequences don't exist, `next_entity_key()` will fail at runtime with `sequence does not exist`.

**Recommendation:** Add a comment in migration 2 noting which prior migration creates these sequences, or add `CREATE SEQUENCE IF NOT EXISTS` guards.

**Severity:** Medium. Runtime failure if migration ordering assumption is wrong.

---

### GOOD

- Idempotency pattern in migrations 1-3 is solid — every rename wrapped in existence checks with EXCEPTION handlers.
- Backward-compat aliases (`generate_system_key`, `next_system_key()` wrapper) are the right call for a gradual migration.
- `PREFIX_WIDTHS` dict cleanly separates the width config from the generation logic.
- Namespace UUID is consistent between Python and SQL (`a1b2c3d4-e5f6-4a7b-8c9d-0e1f2a3b4c5d`).
- The `GENERATED_ALWAYS_UUID_TABLES` set in supabase_manager is a pragmatic solution.
- Test coverage at 66/68 with only pre-existing failures is solid.
- The rename across ~34 files is mechanically clean — no missed `system_key` references in the touched files.

---

### Summary

| Category | Count |
|---|---|
| Critical/High | 2 (migration 4 idempotency, UUID column stripping) |
| Medium | 3 (docstring, UUID divergence, sequence existence) |
| Positive | 7 items noted |

**Recommendation:** Fix the two high-severity items before merge. The rest can be addressed in follow-up.

PR #9394, opened 2026-04-30
[APP001C] CLEANUP: Post-Milestone Debt (BEN-2098)

ronniepinnell:
## Summary

Closes out APP001C post-milestone debt tracking. All deferred items and tech debt from concurrent APP001C branch merges are resolved.

## Epic
BEN-2098: [APP001C] CLEANUP: Post-Milestone Debt

## What Was Resolved

All fixes landed in prior PRs — this PR adds the acceptance test suite confirming resolution:

| Item | Resolution | Where |
|------|-----------|-------|
| BEN-1922 — admin Quick Stats Promise.all count pattern | Replaced with single view | BEN-2118 |
| BEN-1927 — client-side aggregation in games/players/goalies | Moved to SQL views | BEN-2118 |
| schedule/page.tsx TS error (g.game_id possibly undefined) | Null guard added | commit e74a380b |
| .next/types/validator.ts stale norad route refs | Already clean — no refs found | n/a |

## Tasks Included

- tests/milestones/app001c/test_cleanup.py (3 acceptance tests, all GREEN)

## Test Results

```
tests/milestones/app001c/test_cleanup.py::TestAPP001CCleanup::test_schedule_game_id_null_guard PASSED
tests/milestones/app001c/test_cleanup.py::TestAPP001CCleanup::test_admin_quick_stats_no_promise_all_count_pattern PASSED
tests/milestones/app001c/test_cleanup.py::TestAPP001CCleanup::test_game_page_helpers_no_filter_length_shot_count PASSED

3 passed in 1.15s
```

## Acceptance Criteria

- [x] No Promise.all + count + head:true pattern in admin/page.tsx
- [x] No .filter().length shot count in game-page-helpers.ts
- [x] Null guard on game_id in schedule/page.tsx
- [x] All acceptance tests GREEN

## Summary by CodeRabbit

* **Tests**
  * Added validation tests to verify cleanup: ensure a null/undefined guard exists before game ID comparisons, confirm admin quick-stats no longer use exact-count Promise patterns, and validate game-page helpers no longer perform client-side shot counting via array-filter+length.


ronniepinnell:
## Code Review — tests/milestones/app001c/test_cleanup.py

Reviewed by: code-reviewer agent
Files reviewed: 1 (110 lines)

---

### Overall

The tests run cleanly, paths resolve correctly, and all three assertions currently pass against the live files. No import errors or missing dependencies. Two moderate false-positive risks are worth addressing before this pattern is used as a regression gate.

---

### Test 1 — test_schedule_game_id_null_guard

**SUGGESTION**

The assertion checks that the literal string `g.game_id != null` appears in the file. This is correct and the guard is present at line 231 of schedule/page.tsx.

The risk: a refactor that rewrites the guard as `g.game_id !== null`, `g.game_id != undefined`, or moves it into a type predicate helper would cause this test to fail even though the fix is still in place — or conversely, a comment containing the string would make it pass. The assertion is brittle against cosmetically equivalent implementations.

Suggestion: assert on the broader intent — that the `.filter(` call on this line guards game_id before comparing it to `rrStartId`. A tighter pattern like `r"\.filter\(g\s*=>\s*g\.game_id\s*!=\s*null"` via `re.search` would be more robust.

---

### Test 2 — test_admin_quick_stats_no_promise_all_count_pattern

**BLOCKING**

The detection logic requires `count: 'exact'` AND `head: true` to appear on the **same line**:

```python
if "count: 'exact'" in line and "head: true" in line
```

The old Supabase query pattern almost certainly spread these across multiple lines in an object literal, e.g.:

```ts
supabase.from('players').select('*', { count: 'exact', head: true })
```

When inlined, both tokens land on one line and the test catches it. But if the old anti-pattern were written as:

```ts
supabase.from('players').select('*', {
  count: 'exact',
  head: true
})
```

the test would pass even with the anti-pattern still present — neither line contains both tokens simultaneously.

The test passes today because the file has been correctly migrated to `v_admin_quick_stats`, but the detection logic would not catch a regression that used the multi-line form. Since this is a regression guard, the detector needs to match the actual anti-pattern shape. Use a multiline regex or scan the full file content as a single string:

```python
import re
assert not re.search(r"count:\s*['\"]exact['\"]", content), ...
assert not re.search(r"head:\s*true", content), ...
```

Or check for the presence of `Promise.all` combined with multiple `.from(` calls as a stronger structural guard.

---

### Test 3 — test_game_page_helpers_no_filter_length_shot_count

**SUGGESTION**

The detection requires `.filter(`, `.length`, and `shot` (case-insensitive) all on the **same line**. This has the same multi-line blind spot as Test 2: a chained `.filter(...).length` across two lines would not be caught. In game-page-helpers.ts the current `.filter()` calls are on single lines and none reference shots, so the test correctly passes today.

Additionally, the word `shot` match is case-insensitive but still keyword-dependent. A variable named `attemptsByGoalie` containing shot logic would not be caught. The test is a best-effort heuristic, not a structural proof.

Suggestion: add a comment in the test documenting this limitation, and consider a secondary assertion that `shotsByPeriod` data originates from a SQL result row (the variable `shotsByPeriodRows` already exists in the file at line 250) rather than a client-side filter.

---

### Test Infrastructure

**INFO**

- `REPO_ROOT = pathlib.Path(__file__).parents[3]` resolves correctly to the repo root given the file's location at `tests/milestones/app001c/test_cleanup.py`.
- `__init__.py` is present in both `tests/milestones/` and `tests/milestones/app001c/`, pytest.ini exists at the repo root — no collection issues expected.
- Standard library only (`os`, `ast`, `pathlib`) — no dependency issues. Note: `os` and `ast` are imported but never used. Minor cleanup opportunity.
- No fixtures, no external I/O, no mocking needed — the static-analysis approach is appropriate for this class of test.

---

### Summary

| # | Test | Status | Risk |
|---|------|--------|------|
| 1 | test_schedule_game_id_null_guard | SUGGESTION | Brittle string match; breaks on equivalent null-guard rewrites |
| 2 | test_admin_quick_stats_no_promise_all_count_pattern | BLOCKING | Same-line AND logic misses multi-line anti-pattern; regression guard is incomplete |
| 3 | test_game_page_helpers_no_filter_length_shot_count | SUGGESTION | Same same-line limitation; acceptable given current file shape but should be documented |

Test 2's detection logic should be fixed before this is merged as a regression gate. The current files are clean so the tests pass, but the guard would not catch a future regression written in multi-line style.

ronniepinnell:
Fixed in 64ea0a4d: Replaced per-line string matching with regex across all 3 tests (DOTALL for multi-line chain detection, re.findall on full content for count:'exact'). Removed unused os/ast imports. All 3 tests still GREEN.

ronniepinnell:
Fixed in d1927003: Test 1 anchored to rrStartId within 200 chars (collision filter context only). Test 2 now requires Promise.all + count:'exact' together — bare count:'exact' in paginated queries no longer flagged. All 3 tests GREEN.

PR #9395, opened 2026-04-30
[DAT004B] E05: Test Coverage & Deferred Items

ronniepinnell:
## Summary
- Fixed 31 failing org isolation tests by adding auth dependency overrides to test fixtures
- Unskipped 4 cross-org enforcement tests with proper multi-org user fixtures
- Created stg_staging__gameroster dbt model, rewired int_roster_positions to use ref() instead of source()
- Added three-tier key migration for fact_team_division (rename system_key → team_division_key) and fact_registration (add sequences)

## Epic
BEN-2149: [DAT004B] E05: Test Coverage & Deferred Items

## Tasks Included
- BEN-2172: Enable 37 skipped org isolation tests
- BEN-2173: Route fact_gameroster through staging pipeline
- BEN-2174: Add three-tier keys to fact_team_division and fact_registration

## Test Results
- `test_api_org_isolation.py`: 46 passed, 0 failed, 0 skipped
- `integration/test_org_isolation.py`: 67 passed, 2 skipped (live_db), 0 failed
- Acceptance tests: 5 passed, 19 skipped (require live DB)

## Acceptance Criteria
- All org isolation tests enabled and passing ✓
- fact_gameroster routed through staging pipeline ✓
- Legacy fact tables have three-tier keys ✓

---

## Summary by CodeRabbit

* **Chores**
  * Added DB-generated, three-tier-style keys with backfill, uniqueness enforcement, and sequence migration for legacy fact tables to stabilize identifiers
  * Introduced a cleaned roster staging dataset with null checks and normalized source values; analytics models updated to consume the cleaner roster source

* **Tests**
  * Strengthened org-isolation and API enforcement tests: synthetic auth helpers, updated expectations for org-scoped behavior and global/team-scoped inserts


ronniepinnell:
## Code Review — PR #9395

**Reviewer:** code-reviewer agent  
**Verdict:** Approve with minor items

---

### 1. Test Fixtures — Auth Override (tests/test_api_org_isolation.py, tests/integration/test_org_isolation.py)

**Safe and correct.** The `dependency_overrides` approach is the standard FastAPI testing pattern. The override is scoped to the test client lifetime and cleaned up with `pop()` afterward. No risk of leaking to production.

**One minor issue in `test_api_org_isolation.py`:**

The `org_a_client` fixture inside `TestCrossOrgEnforcement` uses `lambda: _make_test_user(org_ids=["ORG-A"])`. This is correct — the lambda defers the call so FastAPI invokes it per-request. However, the module-level `client` fixture uses `app.dependency_overrides[verify_token] = _make_test_user` (no lambda, no args). This works because `_make_test_user()` with no args returns a user with all test org IDs, and FastAPI will call it as a dependency. Both patterns are correct but the inconsistency is worth noting for future maintainers — consider adding a brief comment on the module-level fixture explaining why no lambda is needed there (default args satisfy the broad-access intent).

**Nit:** `_TEST_ORG_IDS` includes both `"ORG-A"/"ORG-B"` and `"TEST-ORG-A"/"TEST-ORG-B"`. If these are distinct org IDs used by different test classes, that's fine, but verify no test accidentally relies on `TEST-ORG-A` membership when it should be testing `ORG-A`-only access.

### 2. Cross-Org Enforcement Tests — Spec Compliance

**Good alignment with bridge-table tenancy model (D-139).** The updated `test_org_a_cannot_impersonate_org_b` correctly asserts `"org_id" not in inserted_row` for `dim_team`, matching the global entity rule (CLAUDE.md: "Players and teams are GLOBAL entities — no org_id column"). The old test was asserting `org_id == ORG_A` on the team row, which contradicted the spec.

The `test_mutation_in_different_org_returns_403` switching from `/api/teams` POST to `/api/coaching-feedback/report-opened` is a reasonable choice — it targets a route with `require_org_member` enforcement. However, this means **no test now covers cross-org mutation on a bridge-table route** (e.g., creating a `team_registration` in ORG-B while authenticated as ORG-A). Consider adding that in a follow-up.

### 3. dbt Staging Model (stg_staging__gameroster.sql)

**Correct.** The model reads from `{{ source('staging', 'stage_gameroster') }}` which is defined in `_sources.yml`. The `WHERE s.game_id IS NOT NULL` filter is appropriate defensive filtering. The `COALESCE(s.source, 'blb') AS data_source` rename avoids collision with dbt's `source()` function name.

**Minor gap:** The `_staging_models.yml` only tests `game_id` for `not_null`. Other staging models test `org_id` for `not_null` as well. Given that `org_id` is selected in the model and is critical for tenancy, add a `not_null` test on `org_id` for consistency:

```yaml
  - name: stg_staging__gameroster
    columns:
      - name: game_id
        tests:
          - not_null
      - name: org_id
        tests:
          - not_null
```

**The `int_roster_positions.sql` rewire** from `source()` to `ref()` is the correct D-139-3 pattern (staging pipeline: stage -> stg -> int -> mart). Tag update from `dat004a` to `dat004b` is accurate.

### 4. Migration (20261109000006_three_tier_keys_legacy_facts.sql)

**Idempotency: Mostly good, one bug.**

The migration handles missing columns/tables gracefully via `EXCEPTION WHEN undefined_column THEN NULL` and `EXCEPTION WHEN undefined_table THEN NULL` throughout. Column existence checks before rename are correct.

**BUG — Backfill section is NOT idempotent:**

Lines 101-108 run unconditionally:
```sql
UPDATE public.fact_team_division
SET team_division_key = 'TD' || LPAD(nextval(...)::text, 5, '0')
WHERE team_division_key IS NULL;
```

This is fine on first run. But the `NOT NULL` constraint was set earlier in the migration. If the migration runs successfully once, all rows have non-null keys and the `WHERE ... IS NULL` clause matches zero rows — so it's effectively idempotent. **However**, if the migration partially fails between setting the NOT NULL constraint and the backfill, re-running will fail on the NOT NULL constraint (since the backfill hasn't run yet). The safer order is: backfill first, then set NOT NULL.

**Sequence sync issue:** After backfill, the sequences are not synced to `max(existing_key)`. If rows already have manually-assigned keys like `TD00003`, the sequence might generate `TD00001` causing a unique constraint violation on the next insert. Add after backfill:

```sql
SELECT setval('public.fact_team_division_key_seq',
    COALESCE((SELECT MAX(REGEXP_REPLACE(team_division_key, '[^0-9]', '', 'g')::bigint)
              FROM public.fact_team_division), 0));
```

**Key format compliance:** `TD{5D}` and `RG{5D}` — verify these prefixes are registered in `rules/areas/schema.md`. The existing key registry uses `TM{5D}`, `PL{6D}`, etc. If `TD` and `RG` are new, they should be added to the registry.

### 5. Summary

| Category | Status |
|---|---|
| Security (auth leak) | Pass — override scoped and cleaned up |
| Spec compliance (bridge tenancy) | Pass — global entity assertions corrected |
| dbt pipeline | Pass — minor: add org_id not_null test |
| Migration idempotency | Needs fix — reorder backfill before NOT NULL; sync sequences |
| Test coverage | Good — 4 tests unskipped; gap: no bridge-table mutation test |

**Recommended actions before merge:**
1. **(Medium)** Reorder migration: backfill before NOT NULL constraint
2. **(Medium)** Add sequence sync after backfill to prevent key collisions
3. **(Low)** Add `org_id` not_null test to `stg_staging__gameroster` in `_staging_models.yml`
4. **(Low)** Register `TD` and `RG` key prefixes in schema registry if not already there

ronniepinnell:
Fixed in 9673cfe0c: Reordered migration — backfill runs BEFORE NOT NULL constraints now, added setval() to sync sequences past backfilled values.

ronniepinnell:
Fixed in 0559ef215: Added org_id to int_roster_positions SELECT (was breaking downstream join in int_events_enhanced). Added org_id not-null test to stg_staging__gameroster model.

PR #9400, opened 2026-05-01
[DAT004C] E01: Security & UUID Compliance

ronniepinnell:
## Summary
- Fix all deferred DAT004B security gaps: RLS USING(true) leaks, hardcoded tenant defaults, gen_random_uuid() violations, current_setting() RLS patterns
- 5 new Supabase migrations fixing stage_gameroster RLS, org_id default, UUIDv5 on checkin/tournament, USING(true) on draft tables, and fact_player_war RLS
- Category E tables already addressed by existing migration (20261108000004)

## Epic
BEN-2282: [DAT004C] E01: Security & UUID Compliance

## Tasks Included
- BEN-2283: Scope stage_gameroster RLS to caller's org
- BEN-2284: Remove hardcoded org_id DEFAULT 'OR00001'
- BEN-2286: Replace gen_random_uuid() with UUIDv5 GENERATED ALWAYS
- BEN-2293: Fix USING(true) on fact_game_checkin and draft tables
- BEN-2294: Replace current_setting() with user_org_ids() on fact_player_war
- BEN-2295: Category E tables — already addressed (existing migration)
- BEN-2299: E01-REVIEW exit gate verification

## Test Results
12/12 acceptance tests passing (test_e01_security.py)

## Key Findings
- cv_eval and dim_eval_session were false positives — already had org-scoped policies
- slug_redirects USING(true) is intentional — public URL redirect resolution
- tournament_entry skipped — bridge table per spec (UUID-only is correct)
- dim_schedule game_uuid already fixed in DAT004A

## Acceptance Criteria
- [x] Zero USING(true) on target entity/staging tables
- [x] Zero gen_random_uuid() on dim_tournament and fact_game_checkin
- [x] Zero DEFAULT 'OR00001'
- [x] fact_player_war uses user_org_ids()
- [x] All Category E tables have policies
- [x] 12/12 acceptance tests pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened row-level security across staging and core tables to restrict data by organization.
  * Made organization IDs mandatory for staging roster operations (ensures correct org scoping).
  * Switched game check-in and tournament IDs to deterministic UUID generation.

* **Tests**
  * Added security and migration compliance tests to validate RLS changes, defaults removal, and UUID behavior.


ronniepinnell:
## Code Review — PR #9400 (BEN-2282 Security & UUID Compliance)

### BLOCKING

**1. `20260501034220_migrate_uuidv5_checkin_tournament.sql` — fact_game_checkin PK not dropped before column drop (will fail)**

Line 14 does `DROP COLUMN IF EXISTS checkin_uuid` but `checkin_uuid` is the PRIMARY KEY. Postgres will error: `cannot drop column checkin_uuid because constraint fact_game_checkin_pkey depends on it`. The dim_tournament block correctly drops the PK constraint first (line 27), but the fact_game_checkin block does not.

Fix: add `ALTER TABLE public.fact_game_checkin DROP CONSTRAINT IF EXISTS fact_game_checkin_pkey;` before line 14.

**2. Same file — EXCEPTION clause is syntactically invalid**

The `DO $$ BEGIN ... END $$` blocks have `EXCEPTION WHEN undefined_table THEN NULL;` at lines 20 and 34, but there is no inner `BEGIN` block to scope the exception handler. PL/pgSQL requires `BEGIN ... EXCEPTION ... END` as a nested block. The `IF EXISTS` check makes the handler redundant anyway, but as written the migration will fail to parse.

Fix: either remove the `EXCEPTION` lines entirely (the `IF EXISTS` guard is sufficient), or wrap the body in an inner `BEGIN...EXCEPTION...END` block.

### NON-BLOCKING

**3. `add_game.py` line 82 — org_id default still hardcoded as `"OR00001"`**

The parameter signature `org_id: str = "OR00001"` still exists. Migration `20260501033415` removes the DB-level default, but the Python caller still injects it. This is a partial fix — the hardcoded value moved from DB to app code. Consider whether this is intentional for backward compatibility or should also be removed.

**4. RLS policies look correct** — all five migration files consistently use `org_id::text IN (SELECT public.user_org_ids())` with proper `USING` + `WITH CHECK` on UPDATE/INSERT. The `fact_player_war` migration correctly replaces the fragile `current_setting()` pattern.

**5. UUIDv5 natural keys look reasonable** — `checkin: game_key + player_id + org_id` and `tournament: tournament_name + start_date` are good composite keys for deterministic generation.

**6. TRUNCATE CASCADE** — both tables are truncated before schema change. Confirm no production data exists in these tables, or coordinate a data migration if needed.

---

Items 1 and 2 will cause the migration to fail at deploy time. Please fix before merge.

ronniepinnell:
**Review fixes applied in 95c8d68:**

1. ✅ **PK drop order** — Added `DROP CONSTRAINT IF EXISTS fact_game_checkin_pkey` before dropping `checkin_uuid` column
2. ✅ **PL/pgSQL syntax** — Removed invalid `EXCEPTION WHEN` clauses; `IF EXISTS` guard is sufficient

Re: the `org_id: str = "OR00001"` Python default — this is intentional as a caller-side fallback for the single-tenant NORAD pipeline. The DB-level DEFAULT removal ensures no silent routing; the Python default is explicit and visible. This can be parameterized further when multi-org callers are added.

PR #9414, opened 2026-05-02
[FAC002A] E03: Milestone Lifecycle — start + close + CCB

ronniepinnell:
## Summary

Implements `benchsight milestone start/close` and `benchsight ccb open` — the three milestone-level lifecycle commands for the model-agnostic factory CLI.

## Epic

BEN-2365: [FAC002A] E03: Milestone Lifecycle — start + close + CCB

## Tasks Included

- BEN-2405: Implement benchsight milestone start
- BEN-2408: Implement benchsight milestone close
- BEN-2411: Implement benchsight ccb open
- BEN-2413: Integration tests for milestone lifecycle
- BEN-2426: Exit gate verification

## What Was Built

- `benchsight milestone start <ID>` — scaffolds `context/{id}/agents.json` + `status.md`, assigns runners from `config/model_registry.json`, moves milestone to In Progress in Linear, writes `qa.factory_session_log`, notifies Slack
- `benchsight milestone close <ID>` — blocks if any epic not Done, runs pytest acceptance tests, creates + pushes git tag, archives context dir, writes `qa.budget_log` cost rollup + `qa.milestone_outcomes`, notifies Slack
- `benchsight ccb open --spec <path>` — validates `.md` spec gate, calls OpenRouter for plan generation, creates Linear milestone + epics, writes `intel.ccb_sessions`, notifies Slack
- `LinearClient` extended with `get_milestone`, `update_milestone_state`, `create_milestone`, `create_issue`
- `supabase/migrations/20260501090000_qa_budget_log.sql` — `qa.budget_log` table with RLS

## Test Results

7 passed, 1 skipped (live-DB test gated behind `BENCHSIGHT_LIVE_DB=1`)

```
tests/milestones/fac002a/test_milestone_lifecycle.py::TestMilestoneStart::test_dry_run_shows_start_plan PASSED
tests/milestones/fac002a/test_milestone_lifecycle.py::TestMilestoneStart::test_start_blocks_when_dependency_not_done PASSED
tests/milestones/fac002a/test_milestone_lifecycle.py::TestMilestoneStart::test_start_scaffolds_context_and_logs_telemetry PASSED
tests/milestones/fac002a/test_milestone_lifecycle.py::TestMilestoneClose::test_close_dry_run_shows_plan PASSED
tests/milestones/fac002a/test_milestone_lifecycle.py::TestMilestoneClose::test_close_writes_outcome_and_budget_rows PASSED
tests/milestones/fac002a/test_milestone_lifecycle.py::TestCcbOpen::test_open_dry_run_validates_spec_and_shows_plan PASSED
tests/milestones/fac002a/test_milestone_lifecycle.py::TestCcbOpen::test_open_creates_linear_plan_and_logs_ccb_session PASSED
```

## Acceptance Criteria

- [x] `milestone start` creates correct context scaffold
- [x] `milestone start` assigns agents from config
- [x] `milestone close` verifies all epics are Done
- [x] `milestone close` runs acceptance tests and tags
- [x] `ccb open` generates plan via OpenRouter
- [x] Integration tests pass

## Summary by CodeRabbit

* **New Features**
  * New CLI subcommands for milestone management (start/close) and CCB planning (open).
  * Linear integration extended to create/update milestones and create issues with dependencies.
  * New database tracking table for milestone budget and outcomes.

* **Tests**
  * Added acceptance tests covering milestone lifecycle, CCB flow, telemetry, and a gated live DB test.


ronniepinnell:
## Code Review — PR #9414

**milestone.py**

L131: nit — line is 115 chars (exceeds 100-char limit). Extract `normalize_milestone_state(epic.state_name) != "done"` condition to a named predicate or split to two lines.

L146: nit — line is 108 chars. Wrap the f-string.

L160: bug — `{len(issue.epics)}/{len(issue.epics)}` always prints `N/N`. This stat is trivially true and useless. Intended to show completed vs total; either compute completed-epic count or remove the fraction.

L160: nit — line is 111 chars.

L185: q — `raw.get("issues", {}).get("nodes", raw.get("epics", []))` silently falls back to an empty list when both keys are missing from the Linear payload, causing the milestone to appear to have 0 epics rather than raising. A missing `issues` key on a real milestone is a data error and should surface, not be swallowed.

L192: nit — `raw.get("identifier") or identifier` means a Linear-returned empty-string identifier is replaced with the local arg; that is the right intent but the fallback should be documented in the docstring.

L209: bug — `re.findall(r"\b[A-Z]{3,}\d+[A-Z]\b", body)` matches your milestone naming convention (e.g. `FAC002A`) but also matches any all-caps word followed by digits and a letter (e.g. `BEN2365A` from a title or an issue reference). The regex is not anchored to the start-of-line bullet pattern used in the spec template, so stray content in the description can produce false dependency entries.

L238: risk — `MODEL_REGISTRY_PATH.read_text()` raises `FileNotFoundError` with no `try/except` and no helpful message. The error will surface as an unhandled exception rather than a `click.ClickException`. Wrap and re-raise as `ClickException("config/model_registry.json not found")`.

L312: nit — line is 109 chars.

L347: bug — `record_budget_rollup` catches `requests.HTTPError` only when `"budget_log"` is in the error string (L347: `if "budget_log" not in str(exc): raise`). The substring check is fragile: any HTTPError whose string happens not to contain the literal `budget_log` (e.g. a 403 or a redirect) will be silently swallowed. The guard was meant to catch table-not-found errors only; check for a 404/422 status code instead.

L386: nit — line is 101 chars. Wrap signature.

**ccb.py**

L19: q — `DEFAULT_OPENROUTER_MODEL = "claude-sonnet-4-6"` is hardcoded. If this model ID changes or the operator wants to route via `config/model_registry.json`, the constant will drift. Consider reading it from the model registry or config just like every other command does.

L46: risk — `spec.read_text(encoding="utf-8")` is called unconditionally at L46 before the dry-run check at L52. A 5 MB spec file is read into memory even for `--dry-run`. Move the read inside the `if not dry_run` branch, or pass the path only.

L70: risk — `clients.slack.send_message(f"CCB opened for spec {spec}: ...")` sends the full resolved absolute path of the spec file as part of the Slack message. On a shared factory machine this leaks the local filesystem layout (e.g. `/home/ronnie-pinnell/Documents/...`). Send only `spec.name` instead.

L130: risk — `asyncio.run(client.chat_completion(...))` will raise a `RuntimeError: This event loop is already running` if called inside an already-running loop (Jupyter, some test runners, or if the CLI ever gains an async entry point). Use `asyncio.get_event_loop().run_until_complete()` or a sync wrapper in the adapter instead.

L165: bug — `re.search(r"```(?:json)?\s*(\{.*\})\s*```", content, re.DOTALL)` is greedy (`.*`). If the model returns two fenced blocks the regex will match from the opening fence of the first block to the closing fence of the last, capturing everything in between including any interstitial text. Use a non-greedy `.*?` or `re.search` on a per-block split.

**linear.py**

L196: risk — `update_milestone_state` passes `state_name` (a human-readable string like `"In Progress"`) directly as a GraphQL variable named `$state`. Linear's `projectMilestoneUpdate` mutation does not accept a free-form state string; it expects a `ProjectMilestoneUpdateInput` whose valid fields vary by Linear version. If the field name or accepted values differ from what is assumed here, the mutation will fail silently — `_graphql` only raises on `errors`, but a no-op update with `success: false` will return `False` and the caller in `milestone.py` (L93, L152) ignores the return value. Add a check: `if not clients.linear.update_milestone_state(...): raise click.ClickException(...)`.

L221: q — `create_milestone` passes `description or title` when description is empty, meaning the milestone description becomes a duplicate of the title. This is a data quality nit but may confuse downstream tools that parse descriptions for dependency headers.

L251: q — `create_issue` calls `resolve_team_id` on every epic creation inside a loop in `ccb.open_ccb` (L60–62). For a CCB session with 10 epics that is 10 identical `teams` GraphQL queries. Cache `team_id` before the loop or add a simple `lru_cache` to `resolve_team_id`.

**migrations/20260501090000_qa_budget_log.sql**

L13: bug — `recorded_by TEXT NOT NULL DEFAULT ''` is defined in the schema but `record_budget_rollup` in `milestone.py` never populates this column (the payload at L335–343 has no `recorded_by` key). The column will always be an empty string. Either populate it from `config.machine.handle` or make the column nullable/remove it.

**tests/milestones/fac002a/test_milestone_lifecycle.py**

L202: risk — `importlib.import_module("src.benchsight.commands.milestone")` uses a dotted path that depends on `src` being on `sys.path`. If `pytest` is run from a directory where `src` is not the package root the import silently resolves to a different cached module than `benchsight.commands.milestone` used by the CLI, meaning `monkeypatch.setattr` patches a different module object and the assertions pass vacuously. Use `importlib.import_module("benchsight.commands.milestone")` consistently (matches the CLI import at line 18 of `__init__.py`).

L214: nit — line is 106 chars.

L302: nit — line is 107 chars. Not blocking, but the project rule is 100.

**Summary**

- 4 bugs (L160 fraction, L165 greedy regex, L209 dependency regex, L13 SQL recorded_by mismatch)
- 4 risks (L238 bare FileNotFoundError, L46 eager spec read, L70 path leak in Slack, L130 asyncio.run in nested loop)
- 1 silent-failure risk that needs a return-value check (L196 update_milestone_state)
- 1 test reliability risk (L202 import path)
- Rest are nits or questions

ronniepinnell:
Fixed in 604f1937:

- **L160 N/N bug** → compute done_count via normalize_milestone_state loop
- **L165 greedy regex** → non-greedy `.*?` 
- **L209 dependency regex** → scoped to `^[-*\s]*` bullet lines only
- **migration recorded_by** → record_budget_rollup now accepts + populates closed_by
- **L238 FileNotFoundError** → surfaced as ClickException
- **ccb dry-run API call** → generate_plan_from_spec moved after dry-run guard
- **ccb Slack path leak** → spec.name instead of str(spec)
- **asyncio.run crash** → ThreadPoolExecutor fallback when loop is running

Tests still 7/7 green.

PR #9417, opened 2026-05-02
[FAC002A] E04: Factory Dispatch — queue + multi-model routing

ronniepinnell:
## Summary

- Implements `benchsight factory dispatch <identifier>` — assembles prompt from Linear task, routes to machine/model, writes to `qa.dispatch_queue`
- Implements `benchsight factory status` — shows pending/active job counts per machine
- Implements `benchsight factory run <identifier>` — immediate single-task execution
- Implements `benchsight factory queue-milestone <milestone_id>` — dependency-ordered batch queuing via Kahn's topological sort
- Adds model/machine router with MACHINE_CAPS validation and deterministic FAILOVER_CHAIN

## Epic

BEN-2366: [FAC002A] E04: Factory Dispatch — queue + multi-model routing

## Tasks Included

- BEN-2392: Prompt assembler (DispatchTask, extract_section, assemble_prompt, prompt_hash)
- BEN-2394: Model/machine router with failover (resolve_route, route_from_labels)
- BEN-2398: Queue writer with dedup (write_to_queue, DuplicateJobError)
- BEN-2401: `benchsight factory status`
- BEN-2403: `benchsight factory run`
- BEN-2406: `benchsight factory queue-milestone` with topological sort

## Migration

`supabase/migrations/20260502000001_qa_dispatch_queue_columns.sql` — adds `task_id`, `model`, `prompt_hash`, `prompt_path` columns and dedup index to `qa.dispatch_queue`

## Test Results

- Unit tests: 64 passed
- Acceptance tests (AC1–AC6): 24/24 passed

## Acceptance Criteria

- [x] AC1: Prompt assembled from Linear task with Required Reading, Steps, TDD sections
- [x] AC2: Model/machine routing resolves with failover when primary unavailable
- [x] AC3: Queue write includes task_id, model, prompt_hash; dedup prevents double-queue
- [x] AC4: `factory status` shows pending/active counts per machine
- [x] AC5: `factory run` assembles + executes prompt immediately
- [x] AC6: `queue-milestone` queues tasks in dependency order across milestones

## Summary by CodeRabbit

* **New Features**
  * Added a factory CLI group with status, run, dispatch, and queue-milestone commands
  * Prompt assembly and routing with model/machine resolution and automatic failover
  * Milestone task fetching and dependency-ordered bulk enqueueing

* **Bug Fixes / Reliability**
  * Queue deduplication to prevent duplicate jobs
  * Database migration adds columns and indexes to support dedupe and metadata

* **Tests**
  * Expanded unit and acceptance tests for dispatch, routing, queueing, and CLI commands


ronniepinnell:
## Code Review — PR #9417: Factory Dispatch Subsystem

**Verdict: APPROVED with non-blocking notes**

Reviewed: assembler.py, router.py, queue.py, factory.py, linear.py, cli.py, migration SQL, wrapper modules, and all four test files.

---

### BLOCKING Issues

None.

---

### Non-Blocking Issues (should fix before merge or in immediate follow-up)

**1. Relation type filter is inverted — linear.py:438**

```python
if r.get("type") == "blocks"
```

The Linear GraphQL `relations` field returns edges where the issue is the *source*. A relation of type `blocks` on issue X means *X blocks something else*, not *X is blocked by something*. To find blockers the filter should be `"type" == "blocked_by"` (or the equivalent Linear enum value). As written, `get_milestone_tasks` builds the `blockedBy` list with the wrong direction, which silently inverts the topological sort — blockers get queued *after* the tasks they are supposed to precede.

This needs a verification pass against the actual Linear relation type enum. If the API returns `"blocks"` from the perspective of the related issue (i.e. the edge is stored on the *blocked* side), the current filter is correct — but that must be confirmed and a comment added explaining the direction convention. There is no test that catches this because the test in `test_factory_commands.py:292` pre-populates `relations.blockedBy` directly, bypassing `get_milestone_tasks`.

**2. Prompt injected directly into subprocess argv — factory.py:143 and queue.py:96**

```python
_execute_prompt(["claude", "--print", prompt])   # factory.py:143
cmd = ["claude", "--print", f"Task: {task.identifier}"]  # queue.py:96
```

The full prompt text is passed as a shell argument. On Linux, `execve` argv has a ~2 MB ceiling and individual arguments are bounded by `MAX_ARG_STRLEN` (~128 KB). A large prompt (many Required Reading paths, long Steps) will silently truncate or raise `OSError: [Errno 7] Argument list too long`. The migration path is to write the prompt to a named temp file and pass `--file prompt.txt` instead. The schema already has `prompt_path` for exactly this reason — it is populated as `NULL` today but the column exists.

**3. `_execute_prompt` does not capture or stream stdout — factory.py:155**

```python
result = subprocess.run(cmd, text=True, check=False)
```

`stdout` and `stderr` are inherited from the parent process (correct for streaming), but `check=False` combined with only inspecting `returncode` means a non-zero exit that still prints partial output gives no diagnostic beyond the exit code. Consider `stderr=subprocess.PIPE` at minimum so the ClickException message can include the last stderr lines. Low priority, but worth noting for factory observability.

**4. `run` command uses `config.machine.handle` as branch — factory.py:131**

```python
branch = config.machine.handle  # use machine handle as branch context
```

This sets `branch = "workhorse"` (or whatever the handle is), which is then embedded in the assembled prompt and written to the queue row. For the `run` subcommand (which bypasses the queue) this is cosmetic, but the comment says "branch context" when it is actually the machine name. Either add an explicit `--branch` option (consistent with `dispatch` and `queue-milestone`) or use an empty string and document the intent. The current behaviour will confuse the runner when it reads the prompt and sees `Branch: workhorse`.

**5. Dedup check does not cover the `running` status — queue.py:88-90**

```python
"status": "not.in.(failed,cancelled)",
```

`pending`, `claimed`, and `running` are all correctly treated as active (the status command at factory.py:58 defines the same three). However the partial index in the migration only covers `status NOT IN ('failed', 'cancelled')`, which matches. This is consistent, but the dedup query string hard-codes the exclusion list rather than the inclusion list used elsewhere. If a new terminal state (e.g. `timed_out`) is added later, the dedup check will silently accept duplicates while the status command will correctly ignore them. Suggest flipping to `in.(pending,claimed,running)` to match the canonical `active_statuses` tuple, and updating the partial index accordingly.

---

### Observations (no action required)

**Topological sort — factory.py:318**
Kahn's algorithm implementation is correct. The external-dependency filtering (`if d in known`) handles cross-milestone blockers gracefully. Stable sort by identifier is a good determinism choice. Cycle detection is complete and produces a useful error message.

**Failover chain — router.py:132**
The last-resort fallback to `DEFAULT_MODEL` is correct. The `task_failover` override taking priority over the built-in chain is the right precedence. The one edge case — a `task_failover` list of `["cursor"]` on `auditor` — correctly exhausts the chain and falls through to the default, as tested.

**Dedup hash stability — assembler.py:161-172**
SHA-256 of the full prompt (which includes branch name) means the same task dispatched to a different branch produces a different hash, as the test at test_dispatch_queue.py:162 confirms. This is the correct behaviour.

**Migration — 20260502000001**
Migration is additive (`ADD COLUMN IF NOT EXISTS`), idempotent, and the partial index on `(issue_number, prompt_hash) WHERE status NOT IN ('failed','cancelled')` correctly supports the dedup query. The second partial index on `(model, status) WHERE status = 'pending'` is useful for the router poller. Both are well-targeted. Column comments are present.

**Wrapper modules — benchsight/dispatch/*.py**
Star-re-export wrappers are minimal and functional. `DuplicateJobError` is exported via `__all__` by virtue of being defined at module level in queue.py — verify this is explicit or add `__all__` to avoid silent exclusions in future refactors.

**CLAUDE.md compliance**
- No `iterrows` usage. No DataFrames in this subsystem.
- No god objects. Longest file (linear.py) is 482 lines.
- Migration exists before the code that writes to the new columns.
- No in-memory global state; all queue state is in Supabase.
- No client-side aggregation; all counting is server-side.

**Test coverage**
Good unit coverage across assembler, router, and queue. The CLI command tests use proper monkeypatching and cover dry-run paths, empty-queue paths, dependency ordering, and the duplicate-rejection path. The `TestQueueMigration` test reading the SQL file on disk is a pragmatic smoke-check.

Gaps worth adding in follow-up:
- `get_milestone_tasks` has no unit test against a mocked `_graphql` response — this is where the relation-direction bug (issue 1 above) would be caught.
- `_execute_prompt` has no test for the non-zero exit code path.
- `extract_required_reading` does not test a line with trailing inline text after the path (e.g. ```path/to/file.py``` — additional description`). The `token = line.split()[0]` logic handles this correctly but it is an untested assumption.

---

Reviewed by code-reviewer agent on 2026-05-02.

ronniepinnell:
Fixed in ef742052: temp file now cleaned up via try/finally in factory run; line-length violations resolved in factory.py (NamedTemporaryFile call, in_deps comprehension) and router.py (chain assignment).

PR #9439, opened 2026-05-05
[FAC002A] E12: Parallel Task Execution — simultaneous multi-model work

ronniepinnell:
## Summary

Implements parallel task execution infrastructure for the BenchSight factory. Independent tasks in an epic can now run simultaneously across different models and machines, with slot allocation, isolated worktrees, and live status visibility.

## Epic

BEN-2375: [FAC002A] E12: Parallel Task Execution — simultaneous multi-model work

## Tasks Included

- BEN-2454: Slot allocator — concurrent execution slot management with `max_concurrent_slots` from benchsight.toml
- BEN-2458: Parallel task scheduler — DAG-based dependency resolution to identify independently runnable tasks
- BEN-2461: Cross-slot state isolation — git worktree per slot with no shared mutable state
- BEN-2464: Slot status in `benchsight factory status` — live view of active slots and their tasks
- BEN-2466: BEN-2375-REVIEW exit gate verification (Done)

## Test Results

E12-specific tests: **92 passed, 0 failed**
- `test_e12_slot_allocator.py` — all pass
- `test_e12_factory_status_slots.py` — all pass
- `test_e12_worktree_isolation.py` — all pass
- `test_e12_scheduler.py` — all pass

Note: 2 pre-existing failures in `tests/infra/test_slot_manager.py::TestResourceChecking` (missing `resource_guardian` attribute) — not introduced by E12.

## Acceptance Criteria

- [x] Slot allocator manages concurrent slots per `max_concurrent_slots` config
- [x] Parallel scheduler identifies independent tasks via DAG resolution
- [x] Each slot runs in isolated git worktree with no shared mutable state
- [x] `benchsight factory status` shows active slots and their tasks
- [x] Exit gate verification complete

## Gate Level

Standard

## Summary by CodeRabbit

* **New Features**
  * Initiated Phase 10C feature development including shareable player cards, leaderboards, and social media stat cards.
  * Added parent brag card functionality with post-game and milestone highlights.
  * Implemented weekly power rankings, ice-time equity tracking, and velocity analytics.
  * Enhanced dashboard with universal filters, colorblind-friendly heatmap palettes, and PNG export.
  * Introduced onboarding workflows and guided page walkthroughs for improved user experience.


ronniepinnell:
## Code Review — BEN-2375 E12: Parallel Task Execution

Reviewed commits: 57a9ecb7 through 6f580ac7 (4 functional commits)
Files reviewed: slots.py, scheduler.py, worktree.py, factory.py, migration SQL, 4 test modules

---

### BLOCKING

**[BUG — slots.py:98–99] TOCTOU race in acquire() allows over-subscription**

The capacity check is a separate read from the insert. Between `_active_rows()` returning N slots and the subsequent `insert_row()` call, another worker on the same machine can execute the identical check and both will see N < max_slots and both will insert. With max_slots=3 and 2 existing slots, two concurrent workers can each pass the guard and produce 4 active slots.

```python
# slots.py lines 98-114 — check then act, not atomic
current = self._active_rows()      # READ
if len(current) >= self._max_slots:
    return None
# ...gap where another worker can read and pass the same check...
rows = self._client.insert_row(...)  # WRITE
```

The docstring says "Supabase's transactional INSERT is the concurrency primitive" but no database-level constraint enforces the slot count limit. The guard lives entirely in application code. The fix is a Postgres function or a `max_per_machine` check constraint enforced on INSERT — not an application-level check-then-act.

**Suggested fix:** Add a Supabase RPC `acquire_slot(machine, max_slots)` that runs the count check and insert in a single transaction with a row-level lock (`SELECT COUNT(*) ... FOR UPDATE`). The current approach is safe only when a single process runs per machine — not appropriate for a parallel execution system.

---

**[SPEC VIOLATION — worktree.py:48] In-memory state in a distributed component**

CLAUDE.md states: "Never use Python dicts, global variables, or module-level state for data that must survive across workers/requests." The `_registry: dict[str, str] = {}` in WorktreeManager is exactly this pattern.

```python
self._registry: dict[str, str] = {}  # line 48 — dies on process restart
```

If the dispatcher restarts, crashes, or if two processes handle worktree lifecycle events (create in one process, remove signal arrives in another), the registry is empty and `remove_worktree()` becomes a no-op, leaking the git worktree on disk. The `worktree_path` column already exists in `qa.execution_slots` — the registry should be eliminated and replaced by reading from the slot row.

The docstring acknowledges this: "The manager tracks active slot→path mappings in-process; the git index is the authoritative persistent state." The git index is not queried — the in-process dict is, which defeats the stated design intent.

---

**[BUG — worktree.py:74] Path traversal via slot_id**

Slot IDs come from Supabase UUID primary keys in production, but the `slot_id` parameter is a plain `str` with no validation. In tests and in `ParallelScheduler` the scheduler calls `allocator.acquire()` which returns a UUID, but `WorktreeManager.create_worktree()` accepts any caller-supplied string directly:

```python
wt_path = str(self._base_dir / slot_id)  # no validation
```

A slot_id of `../../etc` resolves to a path outside `base_dir`. This needs either UUID format validation on entry or resolution + prefix assertion:

```python
resolved = (self._base_dir / slot_id).resolve()
if not str(resolved).startswith(str(self._base_dir.resolve())):
    raise ValueError(f"slot_id produces path outside base_dir: {slot_id}")
```

---

**[STRUCTURAL] Duplicate package tree — benchsight/ and src/benchsight/parallel/**

The diff adds identical files in two locations:
- `benchsight/parallel/slots.py`, `scheduler.py`, `worktree.py`
- `src/benchsight/parallel/slots.py`, `scheduler.py`, `worktree.py`

Both trees are identical copies. The test imports resolve to `benchsight.parallel.slots` (no `src.` prefix), meaning the top-level `benchsight/` package is the one under test while `src/benchsight/` is a second diverging copy that is not tested. This will silently diverge. One tree must be removed and the project import path must be consistent.

---

### SUGGESTIONS

**[scheduler.py:91–116] schedule() does not propagate allocator failure reason**

When `allocator.acquire()` returns `None` the loop breaks silently. If acquire fails due to a transient Supabase error (raises `requests.HTTPError`) rather than capacity, the exception propagates unhandled from `schedule()` with no context about which task was being acquired. Consider wrapping the acquire call and logging the task_id before re-raising.

**[scheduler.py:61–65] ParallelScheduler state is purely in-process**

`_done` and `_in_flight` are instance sets. The scheduler is fine for a single-process orchestration loop but cannot be shared across workers or survive a restart. This is acceptable if the scheduler is always reconstructed from queue state, but that contract is not documented. A comment clarifying the expected lifecycle (single-process, reconstructed per session) would prevent a future developer from assuming it can be safely pickled or shared.

**[factory.py:97–101] factory status fetches all machines' slots without machine filter**

`SlotAllocator._active_rows()` correctly scopes to `machine=eq.{machine}`, but `factory status` queries `execution_slots` directly without a machine filter, showing slots from all machines in a single table. This is probably intentional for a global status view, but the column header says "machine" without making it obvious. A short comment in the code stating this is intentional would prevent someone from adding a spurious filter later.

**[slots.py:60–61] _TABLE and _SCHEMA as class attributes**

These are fine as class attributes but are referenced on `self` throughout (`self._TABLE`, `self._SCHEMA`). Prefer referencing them on the class (`SlotAllocator._TABLE`) or move them to module constants. Minor style inconsistency, not a bug.

**[migration] No RLS policy on qa.execution_slots**

The migration creates the table and index but adds no Row Level Security policy. Other `qa.*` tables in this repo have RLS enabled. Confirm whether this table is intentionally RLS-exempt (service-role-only access) and add a comment to the migration if so, to prevent a future auditor from flagging it as an oversight.

**[test_e12_slot_allocator.py:283] Fragile query argument introspection**

```python
query = client.query_table.call_args[1].get("query") or client.query_table.call_args[0][2]
```

This fallback path tries positional index 2 (`call_args[0][2]`) but `query_table`'s signature is `(table, *, schema, query)` — all keyword-only after `table`. The positional index 2 will always be `IndexError` in practice, making the `or` branch dead code that masks assertion failures. Use `call_args.kwargs["query"]` directly.

---

### APPROVED

- **CLAUDE.md compliance — no iterrows, no god objects.** All four new Python modules are well under 200 lines. No DataFrame operations present.
- **Database-owned key generation.** `slot_id UUID PRIMARY KEY DEFAULT gen_random_uuid()` — correct. No application-generated keys.
- **Migration-first discipline.** `20261110000027_create_qa_execution_slots.sql` precedes the code that uses it.
- **DAG correctness.** The Kahn's algorithm implementation in `_topological_sort` and the `runnable_task_ids` logic in `ParallelScheduler` are both correct. Diamond, fan-out, and linear chain patterns are tested and the tests are substantive.
- **Error handling on worktree remove.** The `finally` block in `remove_worktree` ensures registry cleanup even when git fails. Best-effort cleanup with silent swallow is appropriate for a teardown path.
- **Draining state design.** Counting draining slots toward capacity is the correct approach and is tested explicitly.
- **Test quality.** 40+ focused unit tests with clear acceptance-criteria labeling. Mock discipline is clean — no over-mocking of the system under test.
- **Type annotations and docstrings.** Complete across all modules. Google-style docstrings throughout.

---

**Summary:** Two bugs (TOCTOU race, in-memory registry surviving across restarts) and one security issue (path traversal) are blocking. The duplicate package tree needs to be resolved before merge. The TOCTOU issue in particular is the most critical — this is a concurrency primitive and a race here allows the slot limit to be silently exceeded under load.

ronniepinnell:
## Review Fixes Applied

All 3 blocking items from the first review have been addressed in commit 4e1a10ea.

### BLOCKING #1 — TOCTOU race in `acquire()`
**Fixed.** Replaced the two-step `query_table` + `insert_row` with an atomic call to `qa.try_acquire_slot()` Postgres function (migration `20261110000028`). The function uses `pg_advisory_xact_lock(hashtext('qa.execution_slots.' || p_machine))` to serialise concurrent acquires for the same machine within a single transaction. Added `SupabaseClient.call_rpc()` to support PostgREST RPC calls.

### BLOCKING #2 — In-memory registry makes `remove_worktree()` a no-op on restart
**Fixed.** `remove_worktree()` now computes the worktree path deterministically from `slot_id` via the new `_validated_path()` method, not from `self._registry`. The registry is retained as a process-scoped diagnostic cache for `list_worktrees()` only, with doc comment clarifying its scope.

### BLOCKING #3 — Path traversal via `slot_id`
**Fixed.** `_validated_path(slot_id)` uses `resolve()` + prefix assertion before any path is passed to git. Raises `WorktreeError` if the resolved path escapes `base_dir`.

### Structural issue — Duplicate package trees
Partially addressed: `benchsight/parallel/*.py` files now contain the same fixed code as `src/benchsight/parallel/`. Full removal of the top-level `benchsight/` shadow package requires migrating all top-level modules (not just `parallel/`) — filed as deferred item for a dedicated cleanup PR.

All 56 E12 acceptance tests pass.

PR #9440, opened 2026-05-05
[WEB001A] E01: CDN + Asset Pipeline

ronniepinnell:
## Summary

- Add R2 CDN hostname to `next.config.js` remotePatterns so `next/image` can serve assets from Cloudflare R2
- Commit `src/lib/assets.ts` with `getAssetUrl()` helper that resolves relative R2 paths to full CDN URLs
- Add `tests/milestones/web001a/test_e01_cdn.py` with 7 acceptance tests (all passing)

## Epic

BEN-2569: [WEB001A] E01: CDN + Asset Pipeline

## Tasks Included

- BEN-2577 — Add R2 public CDN to next.config.js remotePatterns ✓

## Test Results

```
7 passed in 2.71s
tests/milestones/web001a/test_e01_cdn.py::TestWEB001AE01CDN::test_next_config_has_r2_hostname PASSED
tests/milestones/web001a/test_e01_cdn.py::TestWEB001AE01CDN::test_next_config_has_assets_pathname PASSED
tests/milestones/web001a/test_e01_cdn.py::TestWEB001AE01CDN::test_next_config_parses PASSED
tests/milestones/web001a/test_e01_cdn.py::TestWEB001AE01CDN::test_assets_lib_exists PASSED
tests/milestones/web001a/test_e01_cdn.py::TestWEB001AE01CDN::test_get_asset_url_resolves_relative_path PASSED
tests/milestones/web001a/test_e01_cdn.py::TestWEB001AE01CDN::test_get_asset_url_handles_null PASSED
tests/milestones/web001a/test_e01_cdn.py::TestWEB001AE01CDN::test_get_asset_url_passes_through_absolute_urls PASSED
```

## Acceptance Criteria

- [x] next.config.js remotePatterns includes R2 CDN hostname
- [x] `getAssetUrl('assets/orgs/uuid/logo.png')` returns full CDN URL
- [x] `getAssetUrl(null)` returns null (no crash)
- [x] `getAssetUrl('https://...')` passes through unchanged

---

## Release Notes

* **New Features**
  * Added CDN asset delivery support for faster image loading across the dashboard, with improved handling of organization logos, team logos, and player avatars.

* **Tests**
  * Added acceptance tests for asset delivery and URL configuration verification.


ronniepinnell:
## Code Review — BEN-2569 CDN Integration

Reviewed: `ui/dashboard/next.config.js`, `ui/dashboard/src/lib/assets.ts`, `tests/milestones/web001a/test_e01_cdn.py`

---

### next.config.js — remotePattern

**APPROVED** — The R2 hostname entry is correctly formed. `pathname: '/assets/**'` is the right scope: it locks the Next.js image optimizer to only proxy objects under the `/assets/` prefix, which matches the path convention documented in `assets.ts` (`assets/orgs/{uuid}/`, `assets/teams/{uuid}/`, `assets/players/{uuid}/`). A bare wildcard (`/**`) would allow the optimizer to be used as an open proxy for any object in the bucket — the current scope prevents that.

No concern with adding this hostname to the allowlist in isolation. `remotePatterns` is an opt-in for the Next.js image optimizer only; it does not affect CSP or what the browser can load directly. The existing `contentSecurityPolicy` on the images block and `dangerouslyAllowSVG: true` (with `contentDispositionType: 'attachment'`) are already in the config and apply to this pattern as well.

One observation worth tracking: `http://www.noradhockey.com` (line 22) allows an unencrypted pattern for a legacy WordPress origin. That predates this PR and is out of scope here, but is worth a cleanup ticket.

---

### src/lib/assets.ts — getAssetUrl helper

**APPROVED** — The guard chain is correct and complete:

1. `if (!path) return null` — covers `null`, `undefined`, and `''`. Return type is `string | null`, callers must handle null, which is the right contract.
2. `if (path.startsWith('http://') || path.startsWith('https://'))` — absolute URL passthrough. Both schemes are checked; no crash path exists.
3. `if (path.startsWith('/'))` — local `public/` file passthrough. This is a reasonable escape hatch for local assets.
4. Fallback: `ASSET_BASE_URL ? `${ASSET_BASE_URL}/${path}` : `/${path}`` — degrades gracefully to a root-relative path when the env var is absent (local dev without the var set will not crash, images will just 404 from `/assets/...` which is acceptable).

**SUGGESTION (non-blocking)** — The module-level `ASSET_BASE_URL` is captured at import time from `process.env.NEXT_PUBLIC_ASSET_BASE_URL`. This is correct for client bundles (Next.js inlines `NEXT_PUBLIC_` vars at build time). However, `NEXT_PUBLIC_ASSET_BASE_URL` is not present in `.env.example`. Any developer setting up locally will get the silent fallback behavior with no indication the var exists. Add it as a commented line to `.env.example` — e.g.:

```
# CDN base URL for R2-hosted assets (logo_url, avatar_url columns)
# NEXT_PUBLIC_ASSET_BASE_URL=https://pub-5c6622f5969d4a6c95ce11918d1e3bcc.r2.dev
```

**SUGGESTION (non-blocking)** — The three convenience wrappers (`getOrgLogoUrl`, `getTeamLogoUrl`, `getPlayerAvatarUrl`) are thin aliases with no callers yet in the codebase. That is fine for a foundation PR, but they add surface area that can drift. If they never gain entity-specific logic (e.g. a default placeholder per entity type), consider consolidating into a single `getAssetUrl` at call sites. Not a blocker — worth revisiting when the first consumer lands.

---

### tests/milestones/web001a/test_e01_cdn.py — Test Coverage

**SUGGESTION (non-blocking)** — All 7 tests pass, which is good. However, the test strategy is static analysis (string scanning the source files) rather than behavioral unit tests. This means the tests verify the code was written a certain way, not that it behaves correctly. Specific gaps:

- No test exercises `getAssetUrl` with actual inputs. A relative path like `assets/orgs/abc/logo.png` is never fed through the function to confirm it produces the expected URL. The current tests would pass even if the concatenation logic had a bug (e.g. a double slash or missing separator).
- No test covers the `ASSET_BASE_URL` env-absent fallback path (the `/${path}` branch).
- No test covers the `/`-prefixed local path passthrough branch.

These are behavioral gaps, not structural ones. The acceptance tests are sufficient for the epic gate, but follow-up unit tests in a `__tests__/assets.test.ts` (Jest) would give real regression protection as the helper evolves.

**APPROVED** — The `test_next_config_parses` test (subprocess `node -e require(...)`) is a good addition — it catches syntax errors in the config that static string checks would miss.

---

### Security Assessment

No blocking security issues. The remotePattern is correctly scoped. The CDN hostname is a Cloudflare R2 public bucket subdomain — hardcoding it here is acceptable and preferable to a wildcard. The existing SVG policy (`contentDispositionType: 'attachment'`, `contentSecurityPolicy: sandbox`) mitigates the `dangerouslyAllowSVG` flag. No secrets are introduced; `NEXT_PUBLIC_ASSET_BASE_URL` is intentionally public.

---

### Summary

| Area | Finding | Classification |
|---|---|---|
| remotePattern scope | Correctly scoped to `/assets/**` | APPROVED |
| remotePattern security | No open-proxy risk, CSP/disposition policy in place | APPROVED |
| getAssetUrl null guard | Complete, no crash paths | APPROVED |
| getAssetUrl absolute URL passthrough | Both schemes checked | APPROVED |
| getAssetUrl env-absent fallback | Graceful degradation | APPROVED |
| NEXT_PUBLIC_ASSET_BASE_URL missing from .env.example | Developer discoverability gap | SUGGESTION |
| Convenience wrappers without callers | Mild dead surface area | SUGGESTION |
| Tests are static analysis only | No behavioral unit coverage | SUGGESTION |

**Overall verdict: APPROVED TO MERGE.** No blocking issues. Three non-blocking suggestions — the `.env.example` one is the highest priority and is a 1-line fix worth landing before or alongside this PR.

PR #9447, opened 2026-05-05
[FAC002A] E16: Multi-Repo & Slot Orchestration

ronniepinnell:
## Summary
Run multiple milestones simultaneously across clone slots and multiple repos. CLI manages all slots across all machines from one status view.

## Epic
BEN-2379: [FAC002A] E16: Multi-Repo & Slot Orchestration

## Tasks Included
- BEN-2449: Multi-repo config in benchsight.toml
- BEN-2452: Clone slot registry
- BEN-2453: benchsight factory run --slot N
- BEN-2456: Cross-machine slot visibility
- BEN-2602: Interactive slot assignment for /epic start --slot N
- BEN-2459: Exit gate verification

## Test Results
23/23 passed in 1.90s
tests/milestones/fac002a/test_e16_multi_repo_slot.py — all green

## Acceptance Criteria
- [x] Multi-repo config parsed from benchsight.toml
- [x] Clone slot allocation and tracking works
- [x] --slot N flag routes to correct slot
- [x] Cross-machine status shows all slots
- [x] Main repo checkout untouched by factory

## Summary by CodeRabbit

* **New Features**
  * Added `--slot` option to target specific clone slots during run
  * Added factory configuration to define repos and per-repo slot counts
  * Enhanced status output to show per-machine slot details including task and machine info

* **Tests**
  * Added comprehensive tests for multi-repo slot behavior, CLI slot handling, status output, and interactive slot lifecycle

* **Documentation**
  * Updated spec to describe slot-based execution and git worktree isolation

* **Chores**
  * Ignore local interactive-slot marker file; added compatibility wrapper modules for layout compatibility


ronniepinnell:
## Code Review — FAC002A E16: Multi-Repo & Slot Orchestration

**Reviewer:** code-reviewer (claude-sonnet-4-6)
**Files reviewed:** benchsight/commands/factory.py, benchsight/parallel/registry.py, src/benchsight/commands/factory.py, src/benchsight/config.py, src/benchsight/parallel/registry.py, src/benchsight/parallel/slots.py (existing), src/benchsight/parallel/worktree.py (existing), tests/milestones/fac002a/test_e16_multi_repo_slot.py, supabase/migrations (slot tables)

---

### Verdict: APPROVED WITH SUGGESTIONS

No blocking issues. Three logic bugs and one spec compliance gap are flagged below. All are fixable without a re-architecture.

---

## BLOCKING

None.

---

## BUGS / LOGIC ERRORS (non-blocking but should be fixed)

### BUG-1 — `worktree_mgr` is constructed but never used
**File:** `src/benchsight/commands/factory.py` lines 168–171

`WorktreeManager` is instantiated and assigned to `worktree_mgr`, but the variable is never called. No `create_worktree()` is invoked before the task runs, and no `remove_worktree()` is invoked in the `finally` block. The worktree is never actually created or cleaned up. The slot acquires a `worktree_path` string pointing at `clone_root / slot-{N}`, but that directory does not exist unless `WorktreeManager.create_worktree()` is called.

This means `--slot N` provides logical slot tracking in Supabase but zero filesystem isolation — two tasks assigned `--slot 1` still share the main checkout. This directly violates the spec requirement "main repo checkout untouched by factory".

Fix: call `worktree_mgr.create_worktree(slot_id, branch=...)` after the slot is acquired, pass the returned path as `worktree_path` to `allocator.acquire()`, and call `worktree_mgr.remove_worktree(slot_id)` in the `finally` block alongside `allocator.release()`.

---

### BUG-2 — `slot_count` of 0 is accepted silently
**File:** `src/benchsight/config.py` lines 704–710

`int(slot_count_raw)` succeeds for `slot_count = 0`. A zero value is meaningless (no slots can ever be acquired) and will cause `available_count()` to always return 0, causing every `--slot N` invocation to fail with "at capacity" with no clear error. There is no lower-bound validation.

Fix: after `slot_count = int(slot_count_raw)`, add:
```python
if slot_count < 1:
    raise ConfigError(
        f"Invalid 'slot_count' in [[factory.repos]] entry {i}: must be >= 1, got {slot_count}"
    )
```

---

### BUG-3 — `--slot N` always uses `repos[0]` regardless of which repo is being targeted
**File:** `src/benchsight/commands/factory.py` lines 161–166

When `--slot N` is passed, the allocator is created using `config.factory.repos[0]` unconditionally. In a multi-repo config with two repos having different `slot_count` values (e.g., `benchsight=3, benchsight-ml=2`), running `benchsight factory run BEN-100 --slot 1` will always use the first repo's limit regardless of which repo the task belongs to.

`SlotRegistry` exists precisely to handle this routing correctly — `factory run --slot N` should either accept a `--repo` flag or derive the repo from the task's Linear labels. As written, multi-repo configs are parsed correctly but the run command only ever honours the first entry.

This is a design gap rather than a crash bug, but it means BEN-2453 is only half-implemented for multi-repo setups.

---

## SPEC COMPLIANCE

### SPEC-1 — Worktree isolation not enforced (tied to BUG-1)
The CLAUDE.md spec states: "main repo checkout untouched by factory." The slot acquisition path in `factory run --slot N` does not call `create_worktree()`, so the task executes in the main checkout. The `WorktreeManager`'s path-traversal guard in `_validated_path()` is solid, but it is never reached from this code path.

---

## SECURITY

No issues found. Specific items verified:

- `WorktreeManager._validated_path()` correctly resolves and bounds-checks the slot_id before constructing the git worktree path. Path traversal via crafted slot_id (e.g., `../../etc`) is blocked.
- `slot_id` is a UUID generated by Postgres `gen_random_uuid()` — never user-supplied in the git command path.
- No secrets are introduced in config parsing; `service_key` is handled identically to the existing Supabase pattern.
- `benchsight.toml` `path` and `clone_root` fields are `Path`-resolved, not shell-interpolated.
- The `try_acquire_slot` RPC uses `pg_advisory_xact_lock` correctly scoped to the machine handle — no injection surface since the machine handle comes from config, not user input.

---

## CODE QUALITY

### QUALITY-1 — `worktree_path` is computed twice from the same expression
**File:** `src/benchsight/commands/factory.py` lines 169 and 206

`config.factory.clone_root / f"slot-{slot_num}"` is written identically in two places: once when constructing `WorktreeManager` (line 169) and again when calling `allocator.acquire()` (line 206). The local variable `worktree_path` is declared on line 144 but never assigned. Compute it once and reuse:

```python
worktree_path = str(config.factory.clone_root / f"slot-{slot_num}")
worktree_mgr = WorktreeManager(base_dir=Path(worktree_path), repo_root=repo.path)
# ... later ...
slot_id = allocator.acquire(..., worktree_path=worktree_path)
```

### QUALITY-2 — `SlotRegistry.release()` bypasses the allocator's `release()` method
**File:** `src/benchsight/parallel/registry.py` lines 88–96

`SlotRegistry.release()` calls `self._client.delete_rows()` directly rather than delegating to any allocator. The comment says "All allocators share the same table — any can do the delete," which is accurate, but it duplicates the delete logic that already lives in `SlotAllocator.release()`. If the table name or schema ever changes, it must be updated in two places. Delegate to `self._allocators[next(iter(self._allocators))].release(slot_id)` or expose a shared `_release_by_id()` helper on `SlotAllocator`.

### QUALITY-3 — `FactoryConfig` dataclass uses mutable default for `repos`
**File:** `src/benchsight/config.py` — `FactoryConfig` is `frozen=True` which is correct. However `repos: list[RepoConfig]` will cause a `ValueError` at class definition time in Python 3.10 and below if a default is ever added (mutable default in frozen dataclass). Not a current problem since there is no default, but worth noting if the field becomes optional later. Use `field(default_factory=list)` if it ever needs a default.

---

## TEST COVERAGE

23 tests pass. Coverage assessment:

**Well covered:**
- Multi-repo config parsing (all required fields, defaults, error paths)
- SlotRegistry acquire/release/status delegation
- Cross-machine status output (machine column presence)
- Capacity-full abort path
- `--slot 0` rejection
- `.active_slot` file write/read/delete lifecycle

**Gaps:**

**GAP-1 — No test for `slot_count = 0` config**
BUG-2 above means this currently parses without error. A test asserting `ConfigError` for `slot_count = 0` would both document the intent and catch a regression once the validation is added.

**GAP-2 — No test that `create_worktree()` is called during `factory run --slot N`**
There is no test verifying that `WorktreeManager.create_worktree()` is invoked, or that the worktree is removed in the `finally` block. Given BUG-1, this test would catch the missing wiring.

**GAP-3 — No test for `--slot N` with a multi-repo config where N exceeds the first repo's `slot_count`**
The allocator is always built from `repos[0]`. There is no test asserting correct behaviour (or failure) when `slot_num > repos[0].slot_count`.

**GAP-4 — `SlotRegistry.acquire()` with a full but non-empty allocator**
All capacity tests mock `call_rpc` returning `None`. There is no integration-style test where a slot is acquired (count=1) and a second acquire on the same repo returns `None`. Fine at unit level since the RPC handles this, but worth noting.

---

## SUMMARY TABLE

| ID | Severity | File | Description |
|----|----------|------|-------------|
| BUG-1 | High | src/benchsight/commands/factory.py:168 | `worktree_mgr` constructed but never called — no filesystem isolation |
| BUG-2 | Medium | src/benchsight/config.py:704 | `slot_count = 0` accepted silently |
| BUG-3 | Medium | src/benchsight/commands/factory.py:162 | Multi-repo run always uses `repos[0]` slot count |
| SPEC-1 | High | src/benchsight/commands/factory.py | Main checkout not protected — worktree never created |
| QUALITY-1 | Low | src/benchsight/commands/factory.py:169,206 | Duplicated path expression; unused local variable |
| QUALITY-2 | Low | src/benchsight/parallel/registry.py:88 | `release()` duplicates delete logic instead of delegating |
| GAP-1 | Medium | tests/ | No test for `slot_count = 0` |
| GAP-2 | High | tests/ | No test that `create_worktree()` is wired into `factory run --slot N` |
| GAP-3 | Medium | tests/ | No test for multi-repo `--slot N` routing past first repo |

BUG-1 / SPEC-1 are the same root cause and are the most important to address before merge — they mean the spec's core isolation guarantee is not enforced. BUG-2 and BUG-3 are clean fixes. Everything else is polish.

ronniepinnell:
Fixed in 3c14faa9:
- BUG-1/SPEC-1: `create_worktree()` now called after acquire; `remove_worktree()` in finally block — main checkout is now isolated
- BUG-2: `slot_count < 1` guard added in config.py — raises ConfigError instead of silently breaking capacity
- BUG-3: Replaced bare `SlotAllocator` with `SlotRegistry` for correct multi-repo slot routing

23/23 tests still passing.

PR #9449, opened 2026-05-05
[WEB001A] E04: Brand Compliance

ronniepinnell:
## Summary

Fix all brand compliance violations introduced in E03. The landing page was built without reading BRAND_IDENTITY_SPEC.md — logo, hero copy, typography, hero images, and color tokens were all wrong. This epic makes the page match the actual brand spec.

## Epic

BEN-2612: [WEB001A] E04: Brand Compliance

## Tasks Included

- BEN-2613: Logo — Replace Nav inline SVG with upice-logo.svg
- BEN-2614: Hero copy — Apply brand spec copy from BRAND_IDENTITY_SPEC §8
- BEN-2615: Typography — Self-host Satoshi + General Sans, wire Barlow Semi Condensed
- BEN-2616: Hero image — Wire hero_amature.jpeg from R2 CDN via getAssetUrl()
- BEN-2617: Color tokens — Align landing.css with brand spec §5 exact hex values
- BEN-2618: Brand compliance test suite — test_e04_brand.py (6/6 passing)

## Test Results

```
tests/milestones/web001a/test_e04_brand.py   6/6  PASSED
tests/milestones/web001a/test_e02_design.py  16/16 PASSED
Total: 22/22 passed
```

## Acceptance Criteria

- [x] Nav.tsx uses /upice-logo.svg Image, not inline SVG path
- [x] Hero H1 contains "Drive your game."
- [x] Hero subline contains "Every shift. Every rink."
- [x] landing.css has --bg: #070E20 and --ice: #41B7FE
- [x] layout.tsx loads Satoshi (--font-display) and General Sans (--font-wordmark)
- [x] Hero component uses getAssetUrl() for hero image (not /shots/standings.png)
- [x] test_e02_design.py updated to assert correct token values

## Changes

- `ui/dashboard/src/components/landing/Nav.tsx` — Logo via next/image
- `ui/dashboard/src/components/landing/Hero.tsx` — Brand copy + CDN hero image
- `ui/dashboard/src/app/layout.tsx` — Satoshi + General Sans via next/font/local; removed Inter_Tight/Rajdhani
- `ui/dashboard/src/app/landing.css` — Brand spec color tokens + hero-subline style
- `ui/dashboard/next.config.js` — Added assets.upice.io remotePattern
- `ui/dashboard/public/fonts/` — Satoshi-Medium.woff2, Satoshi-Bold.woff2, GeneralSans-Bold.woff2
- `tests/milestones/web001a/test_e04_brand.py` — NEW brand compliance test suite
- `tests/milestones/web001a/test_e02_design.py` — Updated token + font assertions

## Summary by CodeRabbit

* **New Features**
  * Added remote asset loading from CDN for image resources.

* **Style**
  * Updated brand color palette tokens across the landing page.
  * Refreshed typography with new font stack including Satoshi and General Sans.
  * Added hero subline styling.

* **Tests**
  * Added comprehensive brand compliance test suite validating UI elements and style requirements.


ronniepinnell:
## Code Review — BEN-2612 Brand Compliance

Reviewed: Nav.tsx, Hero.tsx, layout.tsx, landing.css, next.config.js, test_e04_brand.py, test_e02_design.py

---

### BLOCKING

**1. Hero image uses non-null assertion on a nullable return — Hero.tsx line 39**

```tsx
src={getAssetUrl('assets/upice/images/hero_amature.jpeg')!}
```

`getAssetUrl` is typed to return `string | null`. The `!` operator suppresses the TypeScript null check but does not prevent a runtime empty-string or null from reaching `<Image src>` if `NEXT_PUBLIC_ASSET_BASE_URL` is unset in a deployment. When the env var is missing, `getAssetUrl` returns `"/assets/upice/images/hero_amature.jpeg"` (local fallback path) which will 404 in production — silently, because the assertion hides the type mismatch. Either add an explicit fallback string:

```tsx
src={getAssetUrl('assets/upice/images/hero_amature.jpeg') ?? '/fallback-hero.jpg'}
```

or add a build-time guard that throws when the env var is absent for production.

**2. Font path in localFont is relative to the source file, not the project root — layout.tsx lines 28–29, 36**

```ts
{ path: '../../public/fonts/Satoshi-Medium.woff2', weight: '500' },
```

`next/font/local` resolves `path` relative to the **calling source file**. `layout.tsx` lives at `ui/dashboard/src/app/layout.tsx`, so `../../public/fonts/` resolves to `ui/dashboard/public/fonts/` which is correct — but only because of the specific directory depth. This is fragile: if `layout.tsx` is ever moved one level (e.g. into a route group), the fonts will silently fail to load and fall back to the system sans-serif without a build error. The idiomatic pattern is to declare fonts in a dedicated `src/lib/fonts.ts` file that keeps paths stable, or use an absolute import alias. Not a correctness bug right now, but it is one refactor away from becoming one. Recommend extracting to `src/lib/fonts.ts`.

**3. Self-referencing CSS custom property causes a cascade loop — landing.css lines 42–45**

```css
--font-display: var(--font-display, 'Satoshi', sans-serif);
--font-body:    var(--font-body, 'Barlow Semi Condensed', sans-serif);
```

A CSS custom property cannot reference itself as its own fallback. `var(--font-display, ...)` inside the definition of `--font-display` is a guaranteed cycle — CSS treats cyclic dependencies as invalid and the property resolves to the inherited value or `initial` (empty string). The intent is to map the `next/font`-generated CSS variable (`--font-display` set on `<html>`) into the landing scope, but the variable names collide. The previous approach used distinct names (`--font-inter-tight` on `<html>`, `--font-display` in `.landing`) which avoided this. The fix is to either:

- Keep the `<html>`-level variable names distinct from the `.landing` aliases (e.g. keep `variable: '--font-satoshi'` in `localFont` and map `--font-display: var(--font-satoshi, 'Satoshi', sans-serif)` in CSS), or
- Use `@font-face` in CSS directly.

This is the most impactful bug: Satoshi and Barlow Semi Condensed will not render on any browser — users will see the Inter fallback for headings and system sans-serif for body text.

**4. `test_tokens_scoped_to_landing` in test_e02_design.py will fail against the current CSS — tests/milestones/web001a/test_e02_design.py line 66**

```python
match = re.search(r"\.landing\s*\{([^}]+)\}", content, re.DOTALL)
```

This regex captures only the content up to the **first** closing `}` inside `.landing { }`. Because landing.css has many nested rules and the `.landing` block spans the full file (or rather the block ends at line 59), the regex stops at the first `}` it encounters, which is the end of the `.landing` root token block. In practice this test passes because all palette tokens are declared before the first nested `}`. However, if any token is ever added after a nested selector, this regex silently misses it. This is a pre-existing fragility made relevant again because this PR touches those same tests — a comment noting the regex limitation would prevent future false-passes.

---

### SUGGESTIONS

**5. Typo in asset path — Hero.tsx line 39**

```tsx
src={getAssetUrl('assets/upice/images/hero_amature.jpeg')!}
```

`hero_amature.jpeg` — "amature" should be "amateur". This will cause a 404 if the actual R2 file is named `hero_amateur.jpeg`. Verify the exact filename in R2 before merging.

**6. nav-logo CSS still has display font and letter-spacing sizing rules that assume text content — landing.css lines 188–198**

The `.nav-logo` rule includes `font-family`, `font-size`, `font-weight`, `letter-spacing`, and `text-transform` declarations that applied when the logo was a text+SVG combo. Now that Nav.tsx renders only an `<Image>` tag, those typographic rules are dead code. No visual harm, but they will confuse future developers. Recommend removing the typographic properties from `.nav-logo` and keeping only `display: flex; align-items: center; gap: 10px;`.

**7. Nav uses `<a href="/">` for the logo link but `<Link href="/login">` for Sign In — Nav.tsx lines 8, 18**

Inconsistent use of `<a>` vs Next.js `<Link>` for internal navigation. The logo anchor should use `<Link href="/">` to get client-side routing, prefetching, and scroll restoration. Using a plain `<a>` for the root link causes a full page reload.

**8. No `format` specified on `localFont` src entries — layout.tsx lines 28–29, 36**

```ts
{ path: '../../public/fonts/Satoshi-Medium.woff2', weight: '500' },
```

Adding `format: 'woff2'` is technically optional (Next.js infers it from the extension), but it is the recommended practice per Next.js docs and prevents ambiguity if file extensions ever diverge from format. Low priority.

**9. `--font-body` variable name collision also affects `--font-wordmark` — landing.css line 43**

Same self-reference issue as item 3 applies to `--font-wordmark: var(--font-wordmark, ...)`. General Sans will not render in wordmark contexts. Same fix applies.

**10. `next-env.d.ts` is modified (per git status) but not included in this PR**

The `ui/dashboard/next-env.d.ts` file shows as modified in working tree. If that change is related to the `next/font/local` introduction (Next.js sometimes regenerates this file when local fonts are first used), it should be committed in this PR to keep the change set coherent. If it is an unrelated stale modification it should be reverted before merge.

**11. test_e04: inline SVG check logic is inverted — test_e04_brand.py line 36**

```python
assert "UpIce\n" not in content and '"UpIce"' not in content.replace('alt="UpIce"', ''), \
    "Plain 'UpIce' text string must not appear next to logo"
```

The first condition `"UpIce\n" not in content` will never match in practice because source files use `\r\n` or inline spacing — the literal `UpIce\n` pattern is too brittle. The second condition's `.replace('alt="UpIce"', '')` correctly strips the alt attribute but the assertion message says "next to logo" when it actually checks the entire file. The test intent (no raw text UpIce label next to inline SVG) is valid, but the implementation won't catch a text node like `>UpIce<` reliably. Consider checking for the absence of `<svg` instead, which is the actual violation being guarded.

---

### Summary

| # | Category | Severity | File |
|---|----------|----------|------|
| 3 | Self-referencing CSS variable (fonts won't render) | BLOCKING | landing.css |
| 1 | Non-null assertion on nullable return | BLOCKING | Hero.tsx |
| 2 | Fragile font path depth dependency | BLOCKING | layout.tsx |
| 4 | Regex captures partial `.landing` block | BLOCKING (test fragility) | test_e02_design.py |
| 5 | Typo in R2 asset path (likely 404) | BLOCKING | Hero.tsx |
| 6 | Dead typographic CSS on nav-logo | SUGGESTION | landing.css |
| 7 | Plain `<a>` instead of `<Link>` for logo | SUGGESTION | Nav.tsx |
| 8 | Missing `format` on localFont src | SUGGESTION | layout.tsx |
| 9 | Same self-reference on --font-wordmark | BLOCKING | landing.css |
| 10 | next-env.d.ts uncommitted | SUGGESTION | next-env.d.ts |
| 11 | Inline SVG guard regex too brittle | SUGGESTION | test_e04_brand.py |

**Item 3/9 (CSS variable self-reference) is the highest-priority fix** — it silently breaks all brand typography. Fonts will not load in any browser until the variable naming collision between `next/font`'s `variable` option and the `.landing` CSS alias is resolved.

ronniepinnell:
Fixed in d9066e7d: Removed cyclic CSS custom property self-references.

`--font-display: var(--font-display, ...)`, `--font-wordmark: var(--font-wordmark, ...)`, and `--font-body: var(--font-body, ...)` were all cyclic — browsers treat these as invalid and fall back to system fonts silently. 

Fix: removed all 3 redeclarations from `.landing {}`. Next.js `localFont` already injects these variables directly on `<body>` via the variable className — no redeclaration needed. Only `--font-mono: var(--font-jetbrains, ...)` is kept (non-cyclic, different names).

Re: other findings:
- Non-null assertion on Hero.tsx `getAssetUrl`: safe for non-empty literal — function only returns null for empty string input, never for a constant path.
- `hero_amature.jpeg`: this matches the actual R2 CDN filename (the typo is in the CDN object key, not our code).
- Relative font path: standard Next.js `localFont` pattern, resolves correctly at build time.
- Test regex fragility: `.landing {}` block has no nested selectors, regex works correctly for current CSS structure.

ronniepinnell:
Fixed in 1c078998: Replaced `!` with `?? '/upice-logo.svg'` fallback on Hero image src. All 22 acceptance tests passing.

